Issue #39, opened 2018-11-26
Component CRD spec support ENV vars but not @CompositeAnnotation


cmoulliard:
Component CRD spec support [ENV vars ](https://github.com/snowdrop/component-operator/blob/master/pkg/apis/component/v1alpha1/component_types.go#L51)but not `@CompositeAnnotation`. Even if we specify such info
```
@CompositeApplication(envVars = @Env(name = "key1", value = "val1"))
```
the yaml file generated is empty
```
---
apiVersion: "v1"
kind: "List"
items:
- apiVersion: "v1beta1"
  kind: "Component"
  metadata:
    annotations: {}
    labels: {}
    name: "component-example"
  spec:
    deploymentMode: "innerloop"
    runtime: "spring-boot"
    exposeService: false
    image: []
    env: []
    feature: []
    link: []
    service: []
```

iocanel:
The idea is that an annotation processor may reuse `config` provided by another `processor` in order to avoid dublicating all configuration all across the board. So in this case we are reusing the `KubernetesConfig` that was created by the `KubertnetesAnnotationProcessor`. In the same spirit we are also reusing the configuration provided by the servicecatalog annotation and so on.

So we have the following options here:

### Use `@CompositeApplication` complementary to `@KubernetesApplication` (current approach).

####  Pros:
- Promotes reusability (more lean approach).

#### Cons:
- Might not be obvious.


### Make `@CompositeApplication` stand alone, and replicate everything from `@KubernetesApplication`.

#### Pros
- simplicity in the design
- possibly easier to understand (?)

#### Cons
- code dublication.

PR #40, opened 2018-11-26
Add envVars to CompositeAnnotation

cmoulliard:
- Fix issue #39 
- Add envVars to CompositeAnnotation
- Issues discovered: **EnvVar is added twice - see issue 28** AND **runtime doesn't appear anymore**
```
  spec:
    deploymentMode: "innerloop"
    exposeService: false
    image: []
    env:
    - name: "key1"
      value: "val1"
    - name: "key1"
      value: "val1"
```

### DIFF
--- a/annotations/component-annotations/src/main/java/io/ap4k/component/ComponentHandler.java
+++ b/annotations/component-annotations/src/main/java/io/ap4k/component/ComponentHandler.java
@@ -20,9 +20,11 @@ import io.ap4k.Handler;
 import io.ap4k.Resources;
 import io.ap4k.component.config.CompositeConfig;
 import io.ap4k.component.config.EditableCompositeConfig;
+import io.ap4k.component.decorator.AddEnvToComponent;
 import io.ap4k.component.decorator.AddRuntimeToComponent;
 import io.ap4k.config.ConfigKey;
 import io.ap4k.config.Configuration;
+import io.ap4k.config.Env;
 import io.ap4k.config.KubernetesConfig;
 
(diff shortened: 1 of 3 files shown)


iocanel:
Envrionmnet variables are already supported, with something like this:

     @KubernetesApplication(envVars = @Env(name = "key1", value = "value1"))
     @CompositeApplication
     @SpringBootApplication
     public class Main {

      public static void main(String[] args) {
        SpringApplication.run(Main.class, args);
      }

The idea is that an annotation processor may reuse `config` provided by another `processor` in order to avoid dublicating all configuration all across the board. So in this case we are reusing the `KubernetesConfig` that was created by the `KubertnetesAnnotationProcessor`. In the same spirit we are also reusing the configuration provided by the se rvicecatalog annotation and so on.

So we have the following options here:

# Use `@CompositeApplication` complementary to `@KubernetesApplication` (current approach).

#  Pros:
- Promotes reusability (more lean approach).

# Cons:
- Might not be obvious.


# Make `@CompositeApplication` stand alone, and replicate everything from `@KubernetesApplication`.

## Pros
- simplicity in the design
- possibly easier to understand (?)

## Cons
- code dublication.


@cmoulliard @geoand @metacosm @gytis: wdyt?

geoand:
I would vote for making `@CompositeApplication` stand alone since at this point in the design (being very very early days) I would rather have more flexibility :)

cmoulliard:
Do you want that we create a separate error concerning the second remark `Issues discovered: EnvVar is added twice - see issue 28 AND runtime doesn't appear anymore` @iocanel ?

PR #43, opened 2018-11-27
Isolate processors

iocanel:
This pull move the KubernetesApplication into its own module.
Also makes ComponentApplication standalone (no longer requiring KubernetesApplication to work).


iocanel:
While I was at it I also dealt with a refegression found by @cmoulliard which had to do with custom decorators overwriting each other .

PR #44, opened 2018-11-27
Rename annotations.

iocanel:


annotations/openshift-annotations/src/main/java/io/ap4k/openshift/confg/OpenshiftConfigCustomAdapter.java
```diff
@@ -17,16 +17,16 @@
 package io.ap4k.openshift.confg;
 
 import io.ap4k.openshift.adapt.OpenshiftConfigAdapter;
-import io.ap4k.openshift.annotation.OpenshiftApplication;
+import io.ap4k.openshift.annotation.GenerateOpenhisftManifest;
```
geoand: This doesn't seem right :P `GenerateOpenhisftManifest` -> `GenerateOpenshiftManifest`

iocanel: oooooops

iocanel:
Even though the pull request is ready to be merged, I am having second thoughts about the name `GenerateOpenshiftManifest` etc.

The main reasons are the following:

1. The verb `generate` is very strong and very specific, to the point that it feels limiting.
It would be a great name for a method, but it doesn't feel right for a tool that can be used in numerous ways. One of the soon to add features we are going to add is `modifying exisitng manifests`. When that happens the `GenerateXXXX` will highlight just the 1/2 of the provided functionality.

2. In the case of non executable jars, these annotations should also be used to hint the main method (which is something not obvious with the suggested naming).

I am looking for a name that will tell the user: "You know what? This main class is going to run inside kubernetes."

I will leve this open for a while just in case we get an awesome alternative....

geoand:
I can't really comment since I don't really know what the plans for either `1` or `2` are :P

cmoulliard:
Can you resolve the conflicts please @iocanel ? 
Remark : I think that we must also find a better name ...

PR #53, opened 2018-11-28
fix (#52): Add repackage goal to all spring-boot based examples.

iocanel:


iocanel:
@geoand: The main reason is that I am using the ap4k parent instead. We could possibly add pluginManagement there, but I have mixed feelings about it.

geoand:
@iocanel Understood.

Yeah, I understand your skepticism about `pluginManagement`, especially with such a small configuration item

PR #57, opened 2018-11-28
chore: run integration tests on openshift.

iocanel:


.circleci/config.yml
```diff
@@ -38,4 +38,68 @@ jobs:
 
       # run tests!
       - run: mvn integration-test
-
+  OPENSHIFT_3.10.0:
```
geoand: Actually, we (snowdrop) have a CircleCI orb for Openshift 3.9 and 3.11 that can be used. See [this](https://github.com/snowdrop/component-operator/blob/022a105434f788ad79a74d561056453cc3617692/.circleci/config.yml#L78) or [this](https://github.com/snowdrop/spring-boot-http-booster/blob/a5b271f3bf1356d5e46b75ef7d463ea7d462c1a9/.circleci/config.yml#L13)

Would that be something you would be interested in?
If you need Openshift 3.10 support specifically, we can easily add an orb for that as well :)

iocanel: @geoand: I just copied a config from kubernetes-client. Let me use the orbs.

geoand: Cool

iocanel: @geoand: do u see something obvious?

geoand: Let me check the orbs, but the `dev@master` thing looks suspicious

geoand: Which is not fault of your own of course :). It seems like the code of the generator has not been updated :(

geoand: @iocanel can you try using: `openshift: snowdrop/openshift@0.3.0` please?

iocanel: @geoand: fwiw I also needed to allow uncertified orbs from the circleci project settings. It works now. Thanks a ton!

geoand:
Ah yes, I forgot about that orb enablement! Sorry about that!

Awesome to have integration tests!!!

PR #62, opened 2018-11-30
Add Link annotation

### LINKED ISSUE #55: [feature] Add link annotation for Composite CRD
Add a new `Link` annotation to be used to by the `Composite` Annotation

```
public @interface Link {
  String name();
  String targetcomponentname();
  String kind();
  Env[] envVars() default {};
}
```

cmoulliard:
- Fix #55
- Add Link annotation and fields
- Add decorator to inject Links


cmoulliard:
What should we do now ? @iocanel

PR #69, opened 2018-12-02
feat: initial gradle support.

iocanel:


core/src/main/java/io/ap4k/project/SbtInfoReader.java
```diff
@@ -13,16 +13,51 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.io.IOException;
+import io.ap4k.Ap4kException;
+import java.io.InputStreamReader;
+import java.io.BufferedReader;
+import java.util.stream.Collectors;
+import java.util.Arrays;
 
 public class SbtInfoReader implements BuildInfoReader {
```
cmoulliard: As you added `sbt` tool for Scala language, does that mean that `ap4k` annotations will be supported for Scala projects ? Is it documented ?

readme.md
```diff
@@ -98,6 +98,14 @@ On top of that `lightweight` integration with build tools is provided in order t
 Lightweight integration with build tools, refers to reading information from the build tool config without bringing in the build tool itself into the classpath.
```
cmoulliard: Is it possible to add a concrete example containing `pom.xml` snippet code and what the generated resource will then include coming from the pom.xml's file ?

readme.md
```diff
@@ -98,6 +98,14 @@ On top of that `lightweight` integration with build tools is provided in order t
...
 For example in the case of maven it refers to parsing the pom.xml with DOM in order to fetch the artifactId and version.
 
+Supported build tools:
+
+- maven
+- gradle
+- sbt
+
+For all other build tools, the name and version need to be provided via annotation.
```
cmoulliard: Can you add a table indicating which info is read from which file and where does it fit with one of the annotations ?

readme.md
iocanel: TBH I feel that saying that we just read `artifactId` and `version` that we use as lables and to name resources suffice like images, deployments, service etc suffice.

iocanel:
@cmoulliard: added as much information as possible.

PR #74, opened 2018-12-04
refactor: Clearly name all decorators

geoand:
This is done because when exploring the code, it's not always clear
what the purpose of the class is unless it's clearly named


iocanel:
I have mixed feelings about it.

What I don't like is that the `Decorator` suffix feels redundant given that the class lives in a  package named decorator. 

What I like is that it cleanly separates `Configurators` from `Decorators`. 

Given the api protects you from misusing one in the place of the other (in 99% of cases), I am not sure it worths carrying this suffix all over the place.

geoand:
@iocanel I totally understand your concerns. 
I arrived at this when reading through the code in order to understand what's going on. 
I found it a little confusing when browsing code that the decorator classes didn't have the name and constantly had to go back and forth.

Issue #75, opened 2018-12-06
Separate annotations from processors


iocanel:
Gradle and possibly other tools, have special treatment for artifacts containing annoation processors (e.g. exclude them from the compile path). To reason with gradle its required that the annotation and the processor are in different artifacts and that the processor is explicitly defined as `processor`. 

The best way to achieve this without introducing too much noise is to use shade plugin to split the artifacts and add them to the reactor with different classifiers.

PR #76, opened 2018-12-06
refactor (#75): Use shade plugin to split annotations from processors.

iocanel:


geoand:
Could you please provide some information on the rationale behind this @iocanel ?

iocanel:
@geoand: The problem is described in #75. 

In a few words gradle requires a ceremony to get annotation processors working and it doesn't like having both annotation and processors packaged together. So we are creating a different jar that contains just the annotation package. Also we move some classes to different packages, or we merge classes to make the split easier.

geoand:
Thanks!

I missed the link to the issue initially, sorry about that!

PR #78, opened 2018-12-07
Add test starters and more

iocanel:
This pull request is quite big, so here's an overview of what it includes.

It started as a new feat: `adding junit starters for kubernetes and openshift`.

but it revealed issues in the groovy and gradle land, so it also added fixes around gradle, but also added exampels for gradle and groovy.


examples/spring-boot-with-gradle-on-openshift-example/src/main/groovy/io/ap4k/example/sbonopenshift/Controller.groovy
```diff
@@ -0,0 +1,13 @@
+package io.ap4k.example.sbonopenshift
+
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.bind.annotation.RequestMapping
+
+@RestController
+class Controller {
```
geoand: If we are shooting for the smallest possible example, maybe this should part of the main class?

It's obviously a super minor thing, so feel free to disregard it :P

examples/spring-boot-with-groovy-on-openshift-example/src/main/groovy/io/ap4k/example/sbonopenshift/Controller.groovy
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ *
+**/
+package io.ap4k.example.sbonopenshift
+
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.bind.annotation.RequestMapping
+
+@RestController
+class Controller {
```
geoand: Same as comment above

PR #79, opened 2018-12-07
doc: improve documentation.

iocanel:


readme.md
```diff
@@ -2,41 +2,56 @@
 
 [![CircleCI](https://circleci.com/gh/ap4k/ap4k.svg?style=svg)](https://circleci.com/gh/ap4k/ap4k)
 
-Annotation processor for Kubernetes is a collection of Java annotation processors and tools for generating Kubernetes/Openshift resources at compile time.
+Ap4k is a collection of Java annotations and processors for generating Kubernetes/Openshift manifests at compile time.
```
cmoulliard: Why do you use `manifest` as word instead of `resource` which is the exact technical term ?

iocanel: imho manifest strongly suggests that we are referring to the text representation if the resource. Resource feels a little bit more abstract.

cmoulliard: If you want to keep this word, then I propose to add a few definitions top of the document to explain what is a manifest, resources, ... Make sense @iocanel @metacosm @geoand

geoand: To tell you the truth, I consider the two words interchangeable, but indeed semantically `manifest` is more appropriate

iocanel: @cmoulliard: I prefer using the term that needs less explaining. If this term is resources, then I'll start using that.

wdyt?

cmoulliard: You can use it but then explain what is a manifest and what is generated

iocanel:
@geoand: as promised added some docs on external manifest support. also add tons of doc on examples.

Issue #94, opened 2018-12-19
NPE: when running integration tests on apps without a build enabled.


iocanel:
If `@EnableDockerBuild` or `EnableS2iBuild` are not present the tests need to be skipped, as they are when a cluster is not present.

PR #95, opened 2018-12-19
fix (#94): Tests are now skipped if no build configuration is present.

iocanel:


testing/core-junit/src/main/java/io/ap4k/testing/WithKubernetesConfig.java
```diff
@@ -0,0 +1,47 @@
...
+import io.ap4k.project.Project;
+import io.ap4k.utils.Serialization;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+
+public interface WithKubernetesConfig {
+
+  String KUBERNTES_CONFIG_PATH = "META-INF/ap4k/.config/kubernetes.yml";
```
geoand: This looks like a typo :)

testing/core-junit/src/main/java/io/ap4k/testing/WithPod.java
```diff
@@ -60,27 +60,26 @@ default void injectPod(ExtensionContext context, Object testInstance, Field fiel
...
 
-    namedAnnotation(field).ifPresent(name -> {
-      field.setAccessible(true);
-      try {
-        field.set(testInstance, podForName(context, name));
-      } catch (IllegalAccessException e) {
-        throw Ap4kException.launderThrowable(e);
-      }
-    });
+    String name = namedAnnotation(field).orElseGet(() -> getName());
```
geoand: `orElseGet(() -> getName());` could be `orElseGet(this::getName)`

geoand:
I was thinking about the `hasDockerBuildConfig` condition (and the other relevant conditions).
We should make it clear that when running the tests, it's best to use `mvn clean ....` because otherwise the config could be picked up from previous runs where the flag was enabled even if the current run does not have the flag enabled.

PR #104, opened 2019-01-11
Add support for serviceAccount, workingDir, command and args.

iocanel:


iocanel:
This is broken, I will close and reopen when issues are addressed.

geoand:
So should I ignore the review request for the time being?

core/src/main/java/io/ap4k/kubernetes/decorator/AbstractAddProbeDecorator.java
```diff
@@ -30,15 +30,16 @@
 import java.util.Arrays;
 import java.util.Collections;
 
-public abstract class AbstractAddProbeDecorator extends Decorator<ContainerBuilder> {
+public abstract class AbstractAddProbeDecorator extends ApplicationContainerDecorator {
```
geoand: Could we perhaps have a small comment on what the point of the class is?

geoand:
@iocanel Maybe also update some of the examples to showcase the new features?

core/src/main/java/io/ap4k/kubernetes/decorator/AbstractAddProbeDecorator.java
iocanel: ApplicationContainerDecorator does have a comment.
Added a comment on AbstractAddProbeDecorator.

iocanel:
@geoand: For the time being I don't have a use case common enough that would justify an example.

PR #111, opened 2019-02-04
Decouple ap4k from apt

iocanel:
The idea is that the ap4k should not be bound to apt, although it should reamian the default entry point.

This pull request does the following, moves the `control` logic into a new interface kind called Generator. The generator is added either an Annotation instance, or a Map representations for the annotation, and does the rest.

The existing processosrs are just Generator implementation, that the only things the do is to pass the annotation read to the `Generator` methods.


geoand:
Would it make sense to move break out the apt implementations into separate modules thus leaving the core functionality completely clean of any apt code?

PR #112, opened 2019-02-06
fix: Register hooks when session is closing().

iocanel:
My recent refactoring completely broke all hooks. This is a fix for it.


geoand:
Is there a chance that we could have tests that would catch such breaks in the future?

iocanel:
Good question!

In theory we could use the `maven-invoker-plugin` apprach in order to have `mvn clean install -Dap4k.deploy=true` and then check if that worked. But there are a couple of tricky details we need to take care.

I'd say lets create an issue and have this merged for now.

geoand:
> Good question!
> 
> In theory we could use the `maven-invoker-plugin` apprach in order to have `mvn clean install -Dap4k.deploy=true` and then check if that worked. But there are a couple of tricky details we need to take care.
> 
> I'd say lets create an issue and have this merged for now.

Makes sense!

PR #114, opened 2019-02-06
refactor: Share group, name, version and labels across handlers.

iocanel:
Currently information like:

- group
- name 
- version
- labels 

are only available to Kubernetes and Openshift handlers, as the annotations they handle carry these information. The rest of the hanlders are completely unaware.

Only exception was the source to image hanlder, which used an unintuitive approach that was both confusing to the user and not a solution that can be globally used. As it binds the handler to kubernetes or openshift.

This pull request, moves this information into the resources object which is shared between handlers. Kubernetes/Openshift handlers set the information and the rest just read it. Since handlers are ordered, its guranteed that the info will be set first and read later.


geoand:
CI isn't liking the change :P

core/src/main/java/io/ap4k/Resources.java
```diff
@@ -39,6 +42,14 @@
   private final Map<String, Set<Decorator>> customDecorators = new HashMap<>();
   private final Map<String, KubernetesListBuilder> customGroups = new HashMap<>();
 
+  // The fields below represents info that is meant to be shared across generators.
+  // These are provided by annotations like: KubernetesApplication, OpenshiftApplication etc, but are used by many others.
```
geoand: :+1:

iocanel:
@geoand: ci issues resolved.

PR #118, opened 2019-02-12
Introduce annotation processors for transforming POJOs to CRD

iocanel:
This pull request introduces a new module which pretty much allows us to do pojo -> crd.

This has a number of uses, like:
- integration with jvm-operators
- easy generation of client extensions 

and more.


iocanel:
I think that @Jiri-Kremser has something similar in the the jvm-operators, though I am not sure how exactly that works (is the resource generated compile time? runtime? etc). It would be nice to see if there is room for integration.

annotations/crd-annotations/src/main/java/io/ap4k/crd/util/StringReplace.java
```diff
@@ -0,0 +1,45 @@
...
+package io.ap4k.crd.util;
+
+
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class StringReplace implements Function<String, String> {
+
+    private final String target;
```
geoand: This doesn't need to be a class field since it's only used inside the constructor

annotations/crd-annotations/src/main/java/io/ap4k/crd/util/CamelCase.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class CamelCase  {
+
+  public static final String CAMEL_CASE_PATTERN = "(?<!(^|[A-Z]))(?=[A-Z])|(?<!^)(?=[A-Z][a-z])";
+
+  public static final Function<String, List<String>> SPLIT = s -> Arrays.asList(s.split(CAMEL_CASE_PATTERN));
+  public static final Function<List<String>, String> DASH_JOIN = l -> l.stream().collect(Collectors.joining(","));
```
geoand: Can be replaced by `l -> String.join(",", l);`

annotations/crd-annotations/src/main/java/io/ap4k/crd/util/JsonSchema.java
```diff
@@ -0,0 +1,114 @@
...
+    JSONSchemaPropsBuilder builder = new JSONSchemaPropsBuilder();
+    builder.withType("object");
+    List<String> required =  new ArrayList<>();
+    for (Property property : TypeUtils.allProperties(definition)) {
+      JSONSchemaProps schema = from(property.getTypeRef());
+      if (property.getAnnotations()
+        .stream()
+        .filter(a -> a.getClassRef().getFullyQualifiedName().equals("javax.validation.constraints.NotNull"))
+        .findAny()
+        .isPresent()) {
```
geoand: `filter` + `findAny` + `isPresent` can be replaced by: `anyMatch`

annotations/crd-annotations/src/main/java/io/ap4k/crd/annotation/CustomResource.java
```diff
@@ -0,0 +1,83 @@
...
+   */
+  String shortName() default "";
+  /**
+   * The custom resource version.
+   * @return  The version, or the project version if empty.
+   */
+  String version() default "";
+
+  /**
+   * The resouece scope.
```
geoand: should be `resource`

examples/pojo-to-custom-resource-example/src/main/java/io/ap4k/examples/pojo2crd/Foo.java
```diff
@@ -0,0 +1,13 @@
...
+import io.ap4k.crd.annotation.CustomResource;
+import io.ap4k.crd.confg.Scope;
+
+@CustomResource(group = "ap4k.io", version = "v1", scope = Scope.Namespaced)
+public class Foo {
+  int size;
+  boolean enabled;
+  private Bar bar;
+  private Baz[] bazes;
+  private String[] tags;
```
geoand: Perhaps have a List as well instead of only arrays?

geoand:
Not really an expert in CRD definitions, but this looks very good!

annotations/crd-annotations/src/main/java/io/ap4k/crd/util/JsonSchema.java
iocanel: This is so awesome! I didn't knew that!

geoand: :)

iocanel:
@geoand: shall we merge now?

geoand:
@iocanel I don't see why not :P

PR #129, opened 2019-02-20
feature: make k8s/openshift annotations usable from non-apt builds

geoand:
WDYT @iocanel ?


geoand:
I can also show you how I have been using this if that helps

PR #135, opened 2019-02-22
feat: make more modules usable from no-apt builds

geoand:


annotations/istio-annotations/src/main/java/io/ap4k/istio/generator/DefaultIstioConfigGenerator.java
```diff
@@ -0,0 +1,7 @@
+package io.ap4k.istio.generator;
+
+import javax.annotation.processing.SupportedAnnotationTypes;
+
+@SupportedAnnotationTypes("io.ap4k.istio.annotation.Istio")
```
iocanel: This is a convention used by apt in order to specify annotation classes. I am wondering if it makes for better tooling to just return the list/array of classes as a method.

geoand: I was thinking of that as well.

We can add a method to the generator that returns the supported annotations - that would make sense then what I have done so far.
Would you like me to proceed in that fashion?

geoand: How  about we merge this as is for now and I follow up with a new PR?

PR #143, opened 2019-02-22
refactor: Extract GeneratorOptions into a separate module

geoand:
The fact that the module was in the core meant that it got executed
even when a project was only including the apt versions of annotations


geoand:
Sounds good, change commit right up

PR #148, opened 2019-02-25
feat: session now returns a map to the generated resources

geoand:
Moreover the `SimpleFileWriter` now has the ability to forgo writing the
actual files on the filesystem thus giving the ability for integrations
to control this behavior

### DIFF
--- a/core/src/main/java/io/ap4k/SessionWriter.java
+++ b/core/src/main/java/io/ap4k/SessionWriter.java
@@ -4,6 +4,7 @@ import io.ap4k.deps.kubernetes.api.model.KubernetesList;
 import io.ap4k.kubernetes.config.Configuration;
 import io.ap4k.project.Project;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/ap4k/SessionWriter.java
```diff
@@ -23,32 +24,44 @@
...
+    return result;
   }
 
   /**
    * Writes a {@link Configuration}.
    * @param config  The target session configurations.
+   * @return Map Entry containing the file system path of the written configuration and the actual content as the value
    */
-   void write(Configuration config);
+  Map.Entry<String, String> write(Configuration config);
```
iocanel: Given that the session itself holds such a map, I don't see much point in also returning it here.

The only differennce I see is that one is a Map<String, KubernetesList> and the other is Map<String ,String>. I'd rather have the Session use a Map<String, String> instead.

core/src/main/java/io/ap4k/processor/AptWriter.java
```diff
@@ -85,7 +81,9 @@ public void write(Configuration config) {
       name = name.toLowerCase();
       FileObject yml = processingEnv.getFiler().createResource(StandardLocation.CLASS_OUTPUT, PACKAGE, getProject().getAp4kOutputDir() + "/" + String.format(CONFIG, name, YML));
       try (Writer writer = yml.openWriter()) {
-        writer.write(Serialization.asYaml(config));
+        final String value = Serialization.asYaml(config);
+        writer.write(value);
+        return new AbstractMap.SimpleEntry<>(yml.toString(), value);
```
iocanel: The call to `toString()` kind of scares me.

I think that this is more acurate:
    
    Urls.toFile(yml.toUri().toURL()).getAbsolutePath();

iocanel: Also the javadoc should mention that we are talking about absolute path (we do talk about absolute path, don't we)?

geoand: Yes, that is correct, it's the absolute path

PR #149, opened 2019-02-25
feature: Allow users of @KubernetesApplication to specify a registry

geoand:
This is needed if for example users want to use quay.io to store their
images


geoand:
I'm not convinced yet that this is the best solution, I'm looking into the interaction between `@EnableDockerBuild` and `@KubernetesApplication`

geoand:
Actually I think that having `ApplyRegistryToImageDecorator` apply to `ContainerFluent` would be better. But in that case we have an issue with `ApplyRegistryToImageDecorator` and `ApplyImageDecorator` doing the same thing and the later will override the former.

Perhaps we should have a call @iocanel to discuss?

iocanel:
@geoand: I think that we need to close this one.

PR #152, opened 2019-02-25
fix: issues with map adapters and nested annotations (bump sundrio).

iocanel:


iocanel:
@geoand: Haven't tried but I expect it to fail, with class cast exceptions.

geoand:
:+1

geoand:
Just for the record, I tried running the test with the old sundrio version and it did fail as expected

Issue #153, opened 2019-02-26
@EnableDockerBuild contains redundant info


iocanel:
group, name and version should not be repeated in `@EnableDockerBuild` as they are already included in `@KubernetesApplication` and `@OpenshiftApplication`.

PR #154, opened 2019-02-26
fix (#153): remove group, name and version from EnableDockerConfig.

iocanel:


annotations/docker-annotations/src/main/java/io/ap4k/docker/decorator/ApplyRegistryToImageDecorator.java
```diff
@@ -21,24 +21,28 @@
...
+  private final String registry;
+  private final Resources resources;
 
-  public ApplyRegistryToImageDecorator(String registry, String group, String name, String version) {
-    this.name = name;
-    this.image = Images.getImage(registry, group, name, version);
+  public ApplyRegistryToImageDecorator(Resources resources, String registry) {
+    this.registry=registry;
+    this.resources=resources;
+    String a = "this is shaaait";
```
geoand: I am guessing this was left over by mistake?

annotations/docker-annotations/src/main/java/io/ap4k/docker/decorator/ApplyRegistryToImageDecorator.java
```diff
@@ -21,24 +21,28 @@
...
 import io.ap4k.kubernetes.decorator.Decorator;
 import io.ap4k.utils.Images;
+import io.ap4k.Resources;
 
 public class ApplyRegistryToImageDecorator extends Decorator<DeploymentBuilder> {
 
-  private final String name;
-  private final String image;
+  private final String registry;
+  private final Resources resources;
```
geoand: Does it make sense to use `Coordinates` instead of `Resources`?

iocanel: It felt wrong to pass Resources around while I only needed name, version etc.

annotations/docker-annotations/src/main/java/io/ap4k/docker/decorator/ApplyRegistryToImageDecorator.java
iocanel: LoL

PR #157, opened 2019-03-05
feat: Introduce support for init and sidecar containers.

iocanel:


iocanel:
Tests and examples will be added in a follow up pull request, related to jaeger.

Issue #160, opened 2019-03-06
Add annotations for injecting the jaeger agent to the pod.


iocanel:
Currently, there are two approaches to this:

1. using the opearator

requires setting some labels and the operator will do the dirty work.

2. manual addition for the sidecar.

For both of the above scenarios it would be good to have an annotation that will do the dirty work.

PR #161, opened 2019-03-06
refactor: deployment and container decorators now work regardless of …

iocanel:
…the underlying `kind` (so they can be used for both kubernetes and openshift).

So, this is the prequel to jaeger support (see #160).

The jaeger module will need to be able to inject and customize a sidecar container.

with the additon of sidecars, we will need to be explicit on the container that we are decorating. Also, we need to do it in a platform agnostic manner, so that things work the same for both kubernetes, and openshift.

So, this pull request throws  a litle bit of reflection and generics into the mix, so that decorators can specify the deployment and container they target.

Note: Not all decorators have adopted this logic. There are still decorators that will decorate *ALL* containers (e.g. image pull policy and mount decorators).


geoand:
Tests are complaining :(

iocanel:
@geoand: Tests have been fixed.

core/src/main/java/io/ap4k/kubernetes/decorator/ApplicationContainerDecorator.java
```diff
@@ -1,21 +1,84 @@
...
 /**
  * An abstract class for decorating the application container.
  * This is meant to be used by decorators that are intended to be applied only to the application container (e.g. skip sidecars).
  */
-public abstract class ApplicationContainerDecorator extends Decorator<ContainerFluent<?>>  {
+public abstract class ApplicationContainerDecorator<T> extends Decorator<VisitableBuilder>  {
+
+  /**
+   * For container and deployment name null acts as a wildcards.
+   * Let's use a constant instead, for clarity shake
```
geoand: `clarity` -> `clarity's`

core/src/main/java/io/ap4k/utils/Generics.java
```diff
@@ -0,0 +1,77 @@
...
+
+      Object actualTypeArguments;
+      if (type instanceof Class) {
+        actualTypeArguments = ((Class)type).getTypeParameters();
+      } else {
+        actualTypeArguments = ((ParameterizedType)type).getActualTypeArguments();
+      }
+
+      List<Class> typeArgumentsAsClasses = new ArrayList();
+      Object var12 = actualTypeArguments;
```
geoand: I think we can drop `var12` :P

iocanel: We can't though I do need to copy the orginal code and not what intellij's decompiler makes out of the classes.

geoand: Then let's call it something else :P. `var12 ` gives the generated feeling and I personally don't believe that's a good thing :)

iocanel: Already replaced with the original code.

geoand: :+1:

PR #164, opened 2019-03-08
feat: introduce applicationResourceOutputDir

geoand:
This is done to decouple the location of where the ap4k resources should
be written from where the build places resources.
In most cases the two would be the same, but there are cases (like the
upcoming Quarkus extensions where the two could be different).
This directory will be used later to read application configuration
for use in various Configurator classes


geoand:
I help off from removing the old one as to not break any existing integrations, but I guess that is not really a concern at this point :)

PR #173, opened 2019-03-21
fix (#172): Use the containerPort instead of the hostPort as the rout…

iocanel:
…e targetPort.


geoand:
Maybe we should add an integration test that uses a `Route`?

geoand:
Actually @iocanel would you like to compile a list of nice to have end to end tests so I can start putting them together?

iocanel:
@geoand: The problem with the integration test, is that `host` is cluster specific.

geoand:
Wouldn't the host just `someip.nip.io` for the integration tests?

iocanel:
@geoand: Yeah for integration tests that would do

PR #177, opened 2019-03-22
fix: Rationalize the handling of Route

geoand:
There is no need to have `host` set in order for `Route` to work properly (FMP does not set it either or require the user to supply it)

### DIFF
--- a/examples/source-to-image-example/pom.xml
+++ b/examples/source-to-image-example/pom.xml
@@ -34,6 +34,7 @@ limitations under the License.
     <java.version>1.8</java.version>
 
     <version.spring-boot>2.1.0.RELEASE</version.spring-boot>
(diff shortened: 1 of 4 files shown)


examples/source-to-image-example/pom.xml
```diff
@@ -34,6 +34,7 @@ limitations under the License.
     <java.version>1.8</java.version>
 
     <version.spring-boot>2.1.0.RELEASE</version.spring-boot>
+    <version.awaitability>3.1.0</version.awaitability>
```
Ladicek: `awaitability` -> `awaitility`

geoand: LOLOLOL

geoand: Thanks @Ladicek

PR #185, opened 2019-03-28
feat: add the ability to read the port of a Spring Boot project

geoand:


frameworks/spring-boot/src/main/java/io/ap4k/spring/SpringPropertiesHolder.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+public interface SpringPropertiesHolder extends WithProject {
+
+    AtomicReference<Map<String, Object>> springProperties = new AtomicReference<>(null);
+
+    default Map<String, Object> getSpringProperties() {
+        if (springProperties.get() == null) {
+            final Map<String, Object> properties = new HashMap<>();
+            properties.putAll(getProject().parseAppResourceFile("application.properties"));
```
iocanel: I am wondering if we should also take profiles into consideration.

core/src/main/java/io/ap4k/project/Project.java
```diff
@@ -74,18 +74,26 @@ public Project withAp4kOutputDir(String ap4kOutputDir) {
     return new Project(root, ap4kInputDir, ap4kOutputDir, buildInfo);
   }
 
-  public Map<String, Object> parseAppResourceFile(String resourceName) throws IOException {
```
iocanel: Let's just call it `parseResource` as it could actually parse any resource.

geoand: Fixed

frameworks/spring-boot/src/main/java/io/ap4k/spring/SpringPropertiesHolder.java
geoand: Sorry, I just got notified about this and hadn't seen it.
We could do that yes, *if* the profile is defined in the file and not by some other means (by in my experience that doesn't happen often).

Issue #193, opened 2019-04-17
kubernetes annotations not populated


cmoulliard:
### Issue
The kubernetes annotations are not populated when we use ap4k 0.3-SNAPSHOT and this class + annotations

```
package com.example.demo;

import io.ap4k.component.annotation.CompositeApplication;
import io.ap4k.component.annotation.Link;
import io.ap4k.kubernetes.annotation.Annotation;
import io.ap4k.kubernetes.annotation.Env;
import io.ap4k.kubernetes.annotation.KubernetesApplication;
import io.ap4k.servicecatalog.annotation.Parameter;
import io.ap4k.servicecatalog.annotation.ServiceCatalog;
import io.ap4k.servicecatalog.annotation.ServiceCatalogInstance;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@KubernetesApplication(
        annotations = {
                @Annotation(key = "app.openshift.io/artifact-copy-args", value = "'*.jar'"),
                @Annotation(key = "app.openshift.io/component-name", value = "fruit-backend-sb"),
                @Annotation(key = "app.openshift.io/git-dir", value = "fruit-backend-sb"),
                @Annotation(key = "app.openshift.io/git-ref", value = "master"),
                @Annotation(key = "app.openshift.io/git-uri", value = "'https://github.com/snowdrop/component-operator-demo.git'"),
                @Annotation(key = "app.openshift.io/java-app-jar", value = "fruit-backend-sb-0.0.1-SNAPSHOT.jar"),
                @Annotation(key = "app.openshift.io/runtime-image", value = "fruit-backend-sb")
        })
@CompositeApplication(
        name = "fruit-backend-sb",
        exposeService = true,
        envVars = @Env(
                name = "SPRING_PROFILES_ACTIVE",
                value = "openshift-catalog"),
        links = @Link(
                name = "Secret to be injected as EnvVar using Service's secret",
                targetcomponentname = "fruit-backend-sb",
                kind = "Secret",
                ref = "postgresql-db"))
@ServiceCatalog(
   instances = @ServiceCatalogInstance(
        name = "postgresql-db",
        serviceClass = "dh-postgresql-apb",
        servicePlan = "dev",
        bindingSecret = "postgresql-db",
        parameters = {
                @Parameter(key = "postgresql_user", value = "luke"),
                @Parameter(key = "postgresql_password", value = "secret"),
[...]

PR #194, opened 2019-04-19
fix (#193) annotations and labels are now added globally.

iocanel:


cmoulliard:
/LGTM. Tests passed successfully using Component Operator Demo project

PR #195, opened 2019-04-21
Fix #130: don't mess with groups if empty.

### LINKED ISSUE #130: An empty kubernetes.json was generated without the @KubernetesApplication annotation
This shouldn't have happend! No resource should be generated unless a matching annotation is found.

iocanel:


annotations/kubernetes-annotations/src/it/issue-130/src/test/java/io/ap4k/examples/kuberentes/Issue130Test.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.URL;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+public class Issue130Test {
+
+  @Test
+  public void shouldContainComponent() {
+    URL url =Issue130Test.class.getClassLoader().getResource("META-INF/ap4k/kuberentes.yml");
```
cmoulliard: Should be named `kubernetes.yml`

PR #202, opened 2019-05-07
feat: simplify the definition of http readiness and liveness probes

geoand:
The httpAction is now renamed to httpActionPath where users only need
to set the path


geoand:
@iocanel @metacosm WDYT about this (simplistic) approach?

PR #203, opened 2019-05-08
Merge docker-annotations into kubernetes-annotations.

iocanel:


geoand:
Tests are complaining :(

readme.md
```diff
@@ -679,20 +679,10 @@ It will not generate or customize the docker build in anyway.
...
-```xml
-<dependency>
-  <groupId>io.ap4k</groupId>
-  <artifactId>docker-annotations</artifactId>
-  <version>${project.version}</version>
-</dependency>
-```  
-To enable the hook you need to add the `@EnableDockerBuild` annotation to your main class.    
-Finally, to trigger the hook, you need to pass `-Dap4k.build=true`  as an argument to the build, for example:
+To trigger the hook, you need to pass `-Dap4k.build=true`  as an argument to the build, for example:
```
geoand: Wouldn't `@EnableDockerBuild` a few lines down need to be removed as well?

annotations/kubernetes-annotations/src/main/java/io/ap4k/kubernetes/annotation/KubernetesApplication.java
```diff
@@ -32,7 +33,7 @@
 @Buildable(builderPackage = "io.ap4k.deps.kubernetes.api.builder")
 @Pojo(name = "KubernetesConfig", relativePath = "../config",
       mutable = true,
-      superClass = ApplicationConfiguration.class,
+      superClass = BaseConfig.class,
```
geoand: I don't fully grok this, but are you sure it's correct?

geoand:
Tests still failing unfortunately :(

annotations/kubernetes-annotations/src/main/java/io/ap4k/kubernetes/annotation/KubernetesApplication.java
iocanel: What troubles you? Maybe the name is not as descriptive as it should?

ApplicationConfiguration is the parent of all top level config classes.
BaseConfig is the base for all kubernetes classes (KubernetesConfig, OpenshiftConfig).

geoand: Got it, makes sense now!

geoand:
Tests pass :100:

Issue #171, opened 2019-03-19
Add support for externalizing configuration to property files


iocanel:
It would be nice if the annotation properties can be extenralized to properties/yml files.

Ideally, the externalized configuration will be used with or without annotations.

geoand:
So in this case the annotation processors wouldn't be annotation processors 😁

iocanel:
Yeah, "Not Only Annotation Processors"!

Joking aside, ap4k can already be used without annotations, by calling
generators and feeding them maps/properties.

geoand:
Absolutely, I was the one that forced this feature on you 😂

iocanel:
Fixed

PR #216, opened 2019-05-30
Support configuring ap4k via application.properties

iocanel:


iocanel:
This should address #171

PR #228, opened 2019-06-12
feat: Add prefix to annotationless properties

metacosm:


core/src/main/java/io/ap4k/utils/Maps.java
```diff
@@ -1,37 +1,40 @@
...
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.regex.Pattern;
 
-import io.ap4k.Ap4kException;
-import io.ap4k.deps.jackson.core.type.TypeReference;
-
 public class Maps {
+  private static final String DEKORATE_PREFIX = "dekorate";
```
iocanel: Let's keep it ap4k until we proceed with the renaming.

PR #229, opened 2019-06-13
feat: make component generation work in annotation-less mode

metacosm:


metacosm:
@iocanel I do have a question regarding the co-existence of both annotations and properties: which takes the priority if both are present and, possibly, contradictory?

cmoulliard:
> which takes the priority if both are present and, possibly, contradictory?

@metacosm @iocanel @aureamunoz @geoand 

Can we follow the rules of SpringBoot, Spring where parameters of an annotation can only be changed using application.properties|yaml file ?

Spring Example:
```
@SpringBootApplication
--> Name defined using the following field
spring.application.name=my-spring-application
```

Component Example
```
@ComponentApplication
--> Name of application overriding what we calculate -> artifactId or gradle description
dekorate.component.name=my-fruit-client
dekorate.component.exposeservice=true
```
Remarks:
- That means that we will not pass additional parameters to the Java Annotation !!
- @CompositeApplication should be renamed to @ComponentApplication

geoand:
> > which takes the priority if both are present and, possibly, contradictory?
> 
> @metacosm @iocanel @aureamunoz @geoand
> 
> Can we follow the rules of SpringBoot, Spring where parameters of an annotation can only be changed using application.properties|yaml file ?
> 
> Spring Example:
> 
> ```
> @SpringBootApplication
> --> Name defined using the following field
> spring.application.name=my-spring-application
> ```

Not sure I understand, can you explain a little more please?

cmoulliard:
> Not sure I understand, can you explain a little more please?

This is simple. We currently for the `@CompositeApplication` annotation allows to change the fields name and exposeService, DeploymentMode, ... within the annotation

```
@CompositeApplication(
    name = "fruit-client-sb",
    exposeService = true
)
```

I propose as Spring is doing for many of their annotations, that we only provide an empty annotation and that we use the `application.properties|yaml` file to override the parameters calculated or to add additional infos.

Example
```
@CompositeApplication // Empty Java Annotation
--> what we calculate or let the user to modify the fields within the file
dekorate:
   component:
     name: my-favorite-app // user is changing the name 
     exposervice: false // default is true
     env:
     - name: "foo"
        value: "bar"
```

We could also imagine to completely remove the @ServiceCatalog, @Link annotations as they will be calculated/created by the tools such as kreate and most of the developers will never need to create them manually

Example

```
@Link(
    name = "link-to-fruit-backend",
    componentName = "fruit-client-sb",
    kind = Kind.Env,
    envVars = @Env(
        name = "ENDPOINT_BACKEND",
        value = "http://fruit-backend-sb:8080/api/fruits"
    )
)
```
could be replaced by what the tool will add to the application.properties file
```
dekorate:
    link:
      name: "link-to-fruit-backend"
      componentName: "fruit-client-sb"
      kind: env
      env:
        - name: "ENDPOINT_BACKEND"
           value: "http://fruit-backend-sb:8080/api/fruits"
```

geoand:
Thanks for the explanation.

I haven't really followed the discussions closely, but I don't see why the ability to configure via annotations should be removed for some of the functionality and be properties only.

Also, in my opinion @metacosm properties should take precedence over annotations

cmoulliard:
> Also, in my opinion @metacosm properties should take precedence over annotations

Based on my examples, this is the rule that implicitly I suggest too ;-) (properties should take precedence over annotations)

metacosm:
I agree as well that properties should take the precedence over annotations since it allows for overriding things without having to re-build the project (and it allows for non-Java, third-party tools to more easily interact with `ap4k`).

aureamunoz:
Yes, I agree also with all exposed arguments about priority of properties.

iocanel:
Generally, externalized configuration takes precedence over what's
configured in code.

And we should do the same.

metacosm:
To be clear, though, I don't agree with @cmoulliard's approach of having only empty annotations, just that externalized configuration should be prioritized.

geoand:
Yeah, that's the way I see it as well

annotations/component-annotations/src/main/java/io/ap4k/component/generator/CompositeConfigGenerator.java
```diff
@@ -16,35 +16,39 @@
...
   @Override
   default void add(Element element) {
     CompositeApplication composite = element.getAnnotation(CompositeApplication.class);
-    on(composite != null
+    add(composite != null
        ? new ConfigurationSupplier<>(CompositeConfigAdapter.newBuilder(composite).accept(new ApplyProject(getProject())))
       : new ConfigurationSupplier<>(CompositeConfig.newCompositeConfigBuilder().accept(new ApplyProject(getProject()))));
   }
 
-  default void on(ConfigurationSupplier<CompositeConfig> config) {
```
iocanel: The convention used across generators is that the `add` methods get the initial input and create a supplier out of it and the `on` method just processes the supplier. 

It doesn't really matter as its something strictly internal, I am just mentioning it.

iocanel:
> @iocanel I do have a question regarding the co-existence of both annotations and properties: which takes the priority if both are present and, possibly, contradictory?

There are no guarantees regarding `annotation processor` order.
To deal with this we do have ordering for Handlers and Decorators.

annotations/component-annotations/src/main/java/io/ap4k/component/generator/CompositeConfigGenerator.java
metacosm: @iocanel I actually changed it because it's called `add` in `KubernetesApplicationGenerator`, https://github.com/ap4k/ap4k/blob/master/annotations/kubernetes-annotations/src/main/java/io/ap4k/kubernetes/generator/KubernetesApplicationGenerator.java#L71

Issue #222, opened 2019-06-04
Service YAML Not Generated When Just Using Spring Boot Starter


ryanjbaxter:
When you just include the Spring Boot starter, and do not add `@KubernetesApplication`, no service yaml is generated even though `server.port` is set and there is an `@RestController` present.

You can reproduce this with the spring boot example in the project by removing `@KubernetesApplication`.

iocanel:
I've hit this issue myself and I am currently looking into it.

iocanel:
Fixed.

PR #234, opened 2019-06-26
Add support for WebFlux RouterFunction in spring-boot.

iocanel:


iocanel:
@ryanjbaxter: is this what you had in mind?
Note, this is still affected by #222.

ryanjbaxter:
Yes this covers the non Spring MVC WebFlux use case.  The other use case I can think of is if Spring Cloud Gateway is on the classpath.

iocanel:
@ryanjbaxter: can you provide more info about the case where scg is in the classpath? Should we just treat it like we do with `sck`? Would you like to take a stab at it?

PR #235, opened 2019-06-26
Issue 52 buildspec

aureamunoz:


annotations/component-annotations/src/main/java/io/ap4k/component/handler/ComponentHandler.java
```diff
@@ -17,25 +17,37 @@
...
 import io.ap4k.component.decorator.AddRuntimeTypeToComponentDecorator;
 import io.ap4k.component.decorator.AddRuntimeVersionToComponentDecorator;
 import io.ap4k.component.model.Component;
 import io.ap4k.component.model.ComponentBuilder;
-import io.ap4k.component.model.ComponentFluent;
 import io.ap4k.kubernetes.config.ConfigKey;
 import io.ap4k.kubernetes.config.Configuration;
 import io.ap4k.kubernetes.config.Env;
 import io.ap4k.utils.Strings;
+import org.eclipse.jgit.lib.Repository;
```
iocanel: We want to avoid having any dependency to 3rd party libraries (only exception is compile only dependencies). Everything else is shaded into our big fat `dependencies` jar.

Please add and shade jgit into https://github.com/ap4k/dependencies, and then use it here.

annotations/component-annotations/src/main/java/io/ap4k/component/handler/ComponentHandler.java
```diff
@@ -46,23 +58,38 @@ public int order() {
...
+    return type.equals(ComponentConfig.class) ||
+      type.equals(EditableComponentConfig.class);
   }
 
-  private void addVisitors(CompositeConfig config) {
+  private void addVisitors(ComponentConfig config) {
     String type = config.getAttribute(RUNTIME_TYPE);
     String version = config.getAttribute(RUNTIME_VERSION);
 
+    FileRepositoryBuilder builder = new FileRepositoryBuilder();
```
iocanel: This is awesome! In fact its to awesome to have it limited to the component.
Please create and ScmInfo object with `uri`, `branch` and `commit sha` and have as a property of the Project object. Then we can use it from everywehere.

annotations/component-annotations/src/main/java/io/ap4k/component/handler/ComponentHandler.java
aureamunoz: There is a compilation problem. ap4k-dependencies 0.4-SNAPSHOT uses kubernetes-client 4.2.2 and this class `io.ap4k.deps.kubernetes.api.model.rbac.KubernetesRoleBindingBuilder` used in ap4k is not present in the jar. I have to investigate more.

iocanel: @aureamunoz: no need to investigate. The `Kubernetes` prefix has been removed.

I am about to create a PR and update to latest dependencies.

cmoulliard:
@iocanel Could we merge this PR please ?

PR #239, opened 2019-06-28
chore: bump deps and sundrio.

iocanel:


geoand:
The tests don't seem to like the change

annotations/prometheus-annotations/src/main/java/io/ap4k/prometheus/client/ServiceMonitorHandler.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018 The original authors.
+ * Copyright (C) 2015 Red Hat, Inc.
```
geoand: This doesn't seem right...

iocanel: Fixed

PR #240, opened 2019-07-01
Have KubernetesAnnotationProcessor run when other framework annotations present

ryanjbaxter:
Should address ##222

I imagine we might need to add other annotations as well, just not sure what other ones we need to add

### DIFF
--- a/annotations/kubernetes-annotations/src/main/java/io/ap4k/kubernetes/apt/KubernetesAnnotationProcessor.java
+++ b/annotations/kubernetes-annotations/src/main/java/io/ap4k/kubernetes/apt/KubernetesAnnotationProcessor.java
@@ -31,7 +31,7 @@ import javax.lang.model.element.TypeElement;
 import java.util.Set;
 
 @Description("Generates kubernetes manifests.")
-@SupportedAnnotationTypes("io.ap4k.kubernetes.annotation.KubernetesApplication")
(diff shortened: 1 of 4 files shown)


geoand:
Seems good to me! @iocanel, WDYT ?

iocanel:
This is something that would work, but I am not very comfortable with crossing boundaries and have framework references outside of the framework modules.

Moreover, I think that this is more of a workaround to a specific use case, instead of the actual problem which is about config order. 

My understanding is that #222 is caused because some `configurators` are applied before the actual config is in place. And this is something that can be manifested in a lot of diffrerent ways too (e.g. config provided by annotations and overriden by properties etc).

I'd like to experiement with it a for couple of days, before we can greenlight this PR.

iocanel:
Some more details:

When no annotation is found for one of the available Hanlders (Kubernetes, Openshift and so on) we currently use `handleDefault`which ignores all configuration configurators/visitors. The result is that we ignore everything configured in framework modules.

I am experiementing with applying configuratiors/visitors to the `default` config.

ryanjbaxter:
Thanks @iocanel, I agree, I did not like mixing frameworks into the core kubernetes annotation code....just wasnt sure how else to approach it....I am sure you will find something better :)

iocanel:
@ryanjbaxter I merged #242. If this solves your problem, then we can close this one.

PR #241, opened 2019-07-02
Bump deps sundrio

iocanel:


geoand:
I guess the sundrio artifacts haven't found their way to Maven Central yet?

PR #249, opened 2019-07-09
Combine properties from all YAML and Properties files

ryanjbaxter:
It would be very common to have both an `application` properties file and an `application-kubernetes` properties.  Right now only one can be used.  This would allow us to combine the properties in more than one file.

### DIFF
--- a/core/src/main/java/io/dekorate/processor/AbstractAnnotationProcessor.java
+++ b/core/src/main/java/io/dekorate/processor/AbstractAnnotationProcessor.java
@@ -115,7 +115,7 @@ public abstract class AbstractAnnotationProcessor extends AbstractProcessor impl
 
   private void mergeProperties(Map<String, Object> result, Map<String, Object> newProps) {
     for(String newKey : newProps.keySet()) {
-      if(result.containsKey(newKey)) {
+      if(result.containsKey(newKey) && Map.class.isInstance(result.get(newKey)) && Map.class.isInstance(newProps.get(newKey))) {
         mergeProperties((Map)result.get(newKey), (Map)newProps.get(newKey));
       } else {
         result.putAll(newProps);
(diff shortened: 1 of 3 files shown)


cmoulliard:
LGTM. Many thanks for your PR Ryan. Could this PR offer the possibility in a next release to support application_<profile>.yml files where <profile> is a different profile that we can control using parameter -D-Dspring.profiles.active=<profile> ? @ryanjbaxter

iocanel:
@cmoulliard `-D-Dspring.profiles.active` is a runtime property, while we are generating everything during compile time. We can't possibly know what the user might want.

So, I would go just for `application.properties|yaml` and `application-kubernetes.properties|yaml`.

core/src/main/java/io/dekorate/processor/AbstractAnnotationProcessor.java
```diff
@@ -114,7 +116,15 @@ public synchronized void init(ProcessingEnvironment processingEnv) {
         throw DekorateException.launderThrowable(e);
       } 
     }
-    return Collections.emptyMap();
+    return result;
+  }
+
+  private void mergeKubernetesProperties(Map<String, Object> result, Map<String, Object> newProps) {
+    if(newProps.containsKey("kubernetes") && result.containsKey("kubernetes")) {
```
iocanel: Why are we checking if both maps contain the key `kubernetes`?

I'd like at this point to make no assumptions on they keys as it would be nice if this feature also worked for `openshift` and `component` keys too.

ryanjbaxter: Because if you do `result.putAll(newProps)` even if `newProps` has new `kubernetes` properties everything will get overridden because the property `kubernetes` exists in both `maps` even though the properties under the `map` are different.  So essentially it is a very basic merge functionality.

If there are other keys we should worry about besides `kubernetes` then yes I agree we should add logic for that as well.

I can probably generalize the logic to work for any key.

ryanjbaxter:
@cmoulliard @iocanel it might be nice that the yaml | properties files we look in is customizable in some way, but that is a separate issue I think

Issue #251, opened 2019-07-11
Add contextPath to BuildConfig


gytis:
Charles has added contextPath to the component operator's BuildConfig https://github.com/snowdrop/component-operator/blob/631398984547094dd6440edaf01741754fb01959/pkg/apis/component/v1alpha2/buildconfig_types.go#L16-L18. It needs to be added to Dekorate too.

iocanel:
@gytis @cmoulliard: does it need to be expose to the annotations too?
Also how is that different from the `moduleDirName`?

gytis:
I'll leave the annotations question to Charles, because I'm not yet familiar enough to answer. 

`contextPath` is meant to specify a path to where the `moduleDirName` is located, because `moduleDirName` cannot be a path. E.g. if a module is located in `test/module-a` then `contextPath` would be `test` and `moduleDirName` would be `module-a`. Like in a component-operator repository demo application is under demo directory, while root directory has go code.

cmoulliard:
> Also how is that different from the `moduleDirName`?

Response of Gytis is correct. the `contextPath` corresponds to the s2i parameter `contextPath/Dir` and let the s2i tool to perform a `cd contextPath` before to execute the commands as defined by the s2i assemble script (= maven build for us)

iocanel:
@cmoulliard So I asuume that you also need this exposed via annotations, right?

Also, would you like `dekorate` to try and figure those values out if nothing has been specified?

cmoulliard:
> Also, would you like `dekorate` to try and figure those values out if nothing has been specified?

`contextPath` and `moduleDirName` are dekorate optional fields that our operator will set to their respective default values => `.`

@iocanel @aureamunoz @gytis

iocanel:
@cmoulliard: Given that we have the scm root, the module root paths available, its `trivial` for dekorate to fill in those values, so its pretty doable if you want. In fact I have almost done it.

iocanel:
done.

PR #252, opened 2019-07-11
Add context path to build config

gytis:
https://github.com/dekorateio/dekorate/issues/251.

@iocanel is there any test that I should update for this?


iocanel:
@gytis: No test for that. If you aligned the annotations, to also support contextPath we could add a test or two.

PR #259, opened 2019-07-12
Fix ModuleDirName and ContectPath default values for component operator.

iocanel:


cmoulliard:
LGTM. Is there an integrationTest created supported such features ? @aureamunoz @iocanel

iocanel:
@cmoulliard: Issue52Test does check the acutal module dir name.

cmoulliard:
Then approved.

PR #273, opened 2019-07-16
Some issues

aureamunoz:


annotations/component-annotations/src/it/issue-52/src/main/java/io/dekorate/examples/component/Main.java
```diff
@@ -21,6 +21,7 @@
 import io.dekorate.component.annotation.Link;
 import io.dekorate.component.annotation.ComponentApplication;
 import io.dekorate.kubernetes.annotation.Env;
+import io.dekorate.component.model.DeploymentMode;
```
iocanel: Uneeded

aureamunoz: Ok, I take into account in next modifications :)

Issue #275, opened 2019-07-17
Provide the component.spec.port


cmoulliard:
## Enhancement request

By default, the operator is adding (= hard coding) a value for the service and/or ingress route to be created - `8080`. This information should be communicated according to the framework of the runtime the port to be used using the field

`component.spec.port`

Don't wrap the port number (which is a go int32) with double quotes but instead

```
spec:
    deploymentMode: "dev"
    runtime: "spring-boot"
    version: "2.1.3.RELEASE"
    exposeService: true
    port: 8080
```

geoand:
#278 should fix this

PR #278, opened 2019-07-17
#275

geoand:
feature: add port to Component when it is exposed

fixes: #275

### DIFF
--- a/annotations/component-annotations/src/it/issue-275/src/test/java/io/dekorate/issue275/Issue275Test.java
+++ b/annotations/component-annotations/src/it/issue-275/src/test/java/io/dekorate/issue275/Issue275Test.java
@@ -15,7 +15,7 @@
  * 
 **/
 
-package io.dekorate.issue254;
+package io.dekorate.issue275;
 
 import io.dekorate.deps.kubernetes.api.model.KubernetesList;
 import io.dekorate.deps.kubernetes.api.model.HasMetadata;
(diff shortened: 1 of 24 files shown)


annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
```diff
@@ -110,6 +124,20 @@ private void addVisitors(ComponentConfig config) {
     }
   }
 
+  private KubernetesConfig getKubernetesConfig() {
```
iocanel: We could possibly avoid the dependency to `kubernetes-annotations` by using BaseConfig instead of KubernetesConfig.

However, I don't expect

       configurators.get(BaseConfig.class)


to work. So maybe have Configurators internally iterate and check using `isAssignable` instead of just delegating to a map?

annotations/component-annotations/src/it/issue-275/src/test/java/io/dekorate/issue275/Issue275Test.java
```diff
@@ -0,0 +1,48 @@
...
+import io.dekorate.component.model.Component;
+import io.dekorate.utils.Serialization;
+import java.util.List;
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue254Test {
```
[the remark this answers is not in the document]
geoand: good catch! yes, wrong name :)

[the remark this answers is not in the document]
geoand: well actually, I guess you are viewing an older version, because it is already fixed :)

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
geoand: The good thing is that you have already done that work :wink: 

So it should work OOTB

geoand: But of course when I tried it failed....

iocanel: I have?

iocanel: Wow, just checked the implementation of `get`. I have completly forgoten about it.

geoand: :)

That part seems to work. Something else if failing but I'm tracking it doen

PR #281, opened 2019-07-18
Ensure that the name property of @ComponentApplication is applied

geoand:
Fixes: #276

### DIFF
--- a/annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
+++ b/annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
@@ -42,9 +42,11 @@ import io.dekorate.kubernetes.config.Port;
 import io.dekorate.kubernetes.configurator.ApplyAutoBuild;
 import io.dekorate.project.ApplyProjectInfo;
 import io.dekorate.project.Project;
+import io.dekorate.utils.Labels;
 import io.dekorate.utils.Strings;
 
 import java.nio.file.Path;
+import java.util.Map;
 import java.util.Optional;
 
 public class ComponentHandler implements HandlerFactory, Handler<ComponentConfig>, WithProject {
@@ -150,10 +152,12 @@ public class ComponentHandler implements HandlerFactory, Handler<ComponentConfig
    * @return The component.
    */
   private Component createComponent(ComponentConfig config) {
-       return new ComponentBuilder()
+    Map<String, String> labels = resources.getLabels();
+    labels.put(Labels.APP, config.getName());
+    return new ComponentBuilder()
       .withNewMetadata()
-      .withName(resources.getName())
-      .withLabels(resources.getLabels())
+      .withName(config.getName())
+      .withLabels(labels)
       .endMetadata()
       .withNewSpec()
       .withDeploymentMode(config.getDeploymentMode())
--- a/core/src/main/java/io/dekorate/utils/Labels.java
+++ b/core/src/main/java/io/dekorate/utils/Labels.java
@@ -25,9 +25,9 @@ import io.dekorate.kubernetes.config.Label;
 
 public class Labels {
 
-  private static final String APP = "app";
-  private static final String VERSION = "version";
-  private static final String GROUP = "group";
+  public static final String APP = "app";
+  public static final String VERSION = "version";
+  public static final String GROUP = "group";
 
   /**
    * Creates a {@link Map} with the labels for the {@link BaseConfig}.
--- a/annotations/component-annotations/src/it/issue-276/pom.xml
+++ b/annotations/component-annotations/src/it/issue-276/pom.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>2.1.5.RELEASE</version>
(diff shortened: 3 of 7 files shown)


annotations/component-annotations/src/it/issue-275/src/main/java/io/dekorate/issue275/DemoApplication.java
```diff
@@ -6,7 +6,7 @@
 import io.dekorate.component.annotation.ComponentApplication;
 
 @SpringBootApplication
-@ComponentApplication(exposeService=true)
+@ComponentApplication(exposeService=true, name="customName")
```
iocanel: I am a little bit skeptic about checking for issue #281 inside the #275 integration test.
But definitely not a blocker, more like something to note for future discussions.

geoand: What I have done is update 2 tests - one where we use the custom name and one where the default  (`artifactId` is used)

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
```diff
@@ -152,7 +152,7 @@ private BaseConfig getKubernetesConfig() {
   private Component createComponent(ComponentConfig config) {
        return new ComponentBuilder()
       .withNewMetadata()
-      .withName(resources.getName())
+      .withName(config.getName())
       .withLabels(resources.getLabels())
```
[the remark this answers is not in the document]
geoand: I didn't even know about this :)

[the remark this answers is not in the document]
geoand: I'll go ahead and fix it unless you have a PR ready, in which case I'll close this one

geoand:
On hold for the time being

annotations/component-annotations/src/it/issue-275/src/main/java/io/dekorate/issue275/DemoApplication.java
aureamunoz: Ok, That's what I had done but in different tests so let's take yours ;-)

geoand: I reverted this - there is now a separate test for #276

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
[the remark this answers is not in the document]
geoand: Should be fixes now, thanks for spotting it!

geoand:
Ready for new review

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
```diff
@@ -150,10 +152,12 @@ private BaseConfig getKubernetesConfig() {
    * @return The component.
    */
   private Component createComponent(ComponentConfig config) {
-       return new ComponentBuilder()
+    Map<String, String> labels = resources.getLabels();
+    labels.put(Labels.APP, config.getName());
```
aureamunoz: LGTM but I'm just wondering if it should be done in a decorator? @iocanel

iocanel: I prefer to have as many things as possibly set via decorator, but it's not a blocker for me.

The benefit of using a decoratror is that you don't have to keep track of the origin.

Was the resource created using fallback config? It works.
Was the resource created externally? it works.

geoand: Makes sense. Does it apply to this case though?

iocanel: For the name and labels not, as we don't do that in the rest of the modules.

core/src/main/java/io/dekorate/utils/Labels.java
```diff
@@ -25,7 +25,7 @@
 
 public class Labels {
 
-  private static final String APP = "app";
```
iocanel: Why do we make it public?

geoand: Because it is accessed from `ComponentHandler`

iocanel: I mean I know that you need to access the label from ComponentHanlder, but why don't you use the create labels method?

geoand: Perhaps the best thing to do here would be to extract:

```
    Map<String,String> result =  new HashMap<String, String >() {{
        put(GROUP, config.getGroup());
        put(APP, config.getName());
        put(VERSION, config.getVersion());
      }};
```

into a separate method?

Otherwise calling `createLabels` here will look weird

iocanel: I don't follow. What's weird about calling createLabels?

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
aureamunoz: :-) ok. I agree.

core/src/main/java/io/dekorate/utils/Labels.java
geoand: First of all `ComponentConfig` doesn't extend `BaseConfig`. 
Second, what we want to do here is not to populate the labels of `ComponentConfig`, but instead just add a new label to the `ComponentBuilder`.

Does that make sense? Am I missing something here?

iocanel:
Overall, this seems to me like another side-effect of having the ComponentHandler to run without KuberenetesConfig, which leads to no-one setting the application info.

I am wondering if we could just resolve that using the handler @geoand created. and just do a setApplicationInfo() inside handle like the KubernetesHanlder does.

iocanel:
@geoand ok lets not hold the PR more, just make all constatnts in Labels public to keep the symetry and lets merge.

geoand:
@iocanel done, thanks

PR #282, opened 2019-07-19
feat: make git remote configurable

aureamunoz:


geoand:
I don't really understand how someone would use this, can you please provide an example?

Thanks

aureamunoz:
If you have several 'remotes' in your .git/config, for example:
```
[remote "auri"]
        url = git@github.com:aureamunoz/dekorate.git
        fetch = +refs/heads/*:refs/remotes/auri/*
[remote "origin"]
        url = git@github.com:dekorateio/dekorate.git
        fetch = +refs/heads/*:refs/remotes/origin/*

```
and you may want to pass other than origin, you could do:
`@ComponentApplication(name = "hello-world", remote = "auri", envs = @Env(name = "key1", value = "val1"))`
Is that more clear?

annotations/component-annotations/src/main/java/io/dekorate/component/handler/ComponentHandler.java
```diff
@@ -99,13 +100,17 @@ private void addVisitors(ComponentConfig config) {
...
     if (config.getProject().getScmInfo() != null) {
-      String uri = config.getProject().getScmInfo().getUri();
+      String url = config.getProject().getScmInfo().getUrl();
       String branch = config.getProject().getScmInfo().getBranch();
       String buildTypeType = config.getBuildType();
       Path modulePath = config.getProject().getScmInfo().getRoot().relativize(config.getProject().getRoot());
 
+
+      if (!Strings.isNullOrEmpty(config.getRemote())) {
+        url = Git.getRemoteUrl(config.getProject().getScmInfo().getRoot(), config.getRemote()).map(u -> u.replace(Git.GITHUB_SSH,Git.GITHUB_HTTPS)).orElse(url);
```
geoand: This statement is a little to compact for my taste, bu that's just me :)

annotations/component-annotations/src/main/java/io/dekorate/component/annotation/ComponentApplication.java
```diff
@@ -44,4 +44,6 @@
   Env[] envs() default {};
 
   String buildType() default "s2i";
+
+  String remote() default "";
```
iocanel: Default value should be origin.

aureamunoz: Default value is 'origin' as when Scm info is built origin is setted:
`String url = Git.getRemoteUrl(path, **Git.ORIGIN**).map(u -> u.replace(Git.GITHUB_SSH, Git.GITHUB_HTTPS)).orElse(null);`
If it's expressed at annotation level, in the handler it will never be empty and hard to know if it should be overrided. Am I wrong?

iocanel: Not sure why that's a problem. Instead of checking for string being empty you can check for string being equal to "origin".

It's no biggie, but its always better if defaults are clearly expressed. Easier for the user to understand what's going on...

aureamunoz: Yes, true, I didn't see it that way but yes, it's better like this :-)

Issue #283, opened 2019-07-20
Document how to use Dekorate in conjunction with other annotation processors


noelo:

iocanel:
AFAIK, nothing special is required.

In dekorate itself we use often 2 even 3 processors, without any special config.
And this had been my experience with other processors too.

What kind of issue are you having?

noelo:
The dekorate processors were not being used unless I added them to the build plugin in the pom. The original pom is at https://github.com/redhat-cop/pathfinder/blob/master/pathfinder-server/pom.xml without the dekorate processor. This code was originally built using JHipster which we're slowly removing. I wonder if the dekorate processors are not being used as there are other processors explicitly configured.

noelo:
I just tried without any processors configured in the compiler-plugin and it works as it should with dekorate being invoked.

iocanel:
@noelo: so the problem was that you either need to be explicit about all processors or no processor at all right?

noelo:
yes, that's what I'm seeing so far

iocanel:
@noelo Thats also my understanding so, why don't we close this issue and rework the pull request so that it highlights how to add the dekorate dependencies via compiler plugin configuration and its benefits. wdyt?

noelo:
Sure, will do

PR #284, opened 2019-07-20
Fix for interation with other annotation processors #283

noelo:


iocanel:
following up our discussion in #283 I think that there is no issue in using dekorate along with other processors. 

However, I see value in documenting the approach on adding the processors as part of the compiler plugin and its benefits.

noelo:
Great, just checking do you need me to do anything with this ?

Issue #256, opened 2019-07-12
Have a clean separation between examples and integration tests


iocanel:
The idea behind the examples modules was to demonstrate common use cases that can be used as recipes by the users.

Overtime, some of the examples have been modified to act as integration tests and started adding things that no longer make sense for the original use case. The result is examples that are complex and ugly. 

As we have `maven-invoker-plugin` tests for regressions and integration testing I would like to have a clean separation between the two.

This issue is about cleaning examples up and moving anything that doesn't belong there to integration tests.

iocanel:
Went though some of our examples and feel that we need to revisit.

PR #291, opened 2019-07-25
refactor: separation between examples and integration tests renaming

aureamunoz:
Related to issue #256
Here's what I did (I hope this helps)

- I've renamed integration tests within `component-annotation/it` in order to have more details about what it checks in the names.
- The old `/examples/component-example` actually was, in my opinion, an integration test for the overriding of annotation config by values setted in a `application.yml` file so I moved this to `component-annotations/src/it/component-feat-229-override-annotationbased-config`
- The new `/examples/component-example` contains the simplest use case for `@ComponentAppliciation` without any specifique config so it checks the default values.


examples/component-example/src/main/java/io/dekorate/examples/component/Main.java
```diff
@@ -22,13 +22,10 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@ComponentApplication(name = "hello-spring-boot", exposeService = true, envs = @Env(name = "key1", value = "val1"))
-@Link(name = "should-be-overridden-by-properties", componentName = "should-be-overridden-by-properties")
-@SpringBootApplication
```
iocanel: I would keep spring, as its a nice way of show casing how the `runtime`, `version` properties of component are popullated.

The rest I agree that should go.

annotations/component-annotations/src/it/component-feat-275-provide-port/pom.xml
```diff
@@ -9,7 +9,7 @@
 		<relativePath/> <!-- lookup parent from repository -->
 	</parent>
 	<groupId>io.dekorate</groupId>
-	<artifactId>issue-275</artifactId>
+	<artifactId>component-feat-275-provide-port</artifactId>
```
iocanel: Don't have versy strong feelings about it, but I do feel that `component-` is a bit redundant, given that the tests live under `component-annoations`.

examples/component-example/src/main/java/io/dekorate/examples/component/Main.java
aureamunoz: Ok, I agree, I hadn't thought about it. I'm gonna keep spring :-)

annotations/component-annotations/src/it/component-feat-275-provide-port/pom.xml
aureamunoz: Yes, good remark. I didn't like it completely but I didn't know why.

PR #307, opened 2019-08-13
Update to halkyon

metacosm:


aureamunoz:
LGTM. After the fix of checks failed...

PR #312, opened 2019-08-19
doc: add some doc about annotation-less usage

aureamunoz:
Related to issue #304

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -615,7 +615,7 @@ The order where these resources are used are:
 
 Here's the full list of supported [configuration options](config.md).
 
-####### Examples
+###### Examples
 You can find [here](https://github.com/dekorateio/dekorate/blob/master/examples/halkyon-example-annotationless-properties/src/main/resources/application.properties) and [here](https://github.com/dekorateio/dekorate/blob/master/annotations/halkyon-annotations/src/it/feat-229-override-annotationbased-config/src/main/resources/application.yml) some examples of usage. Note that some items does not end up in the same hierarchical place where they were configured. Let's take an example:
 
 the following deployment mode configuration:
--- a/readme.md
+++ b/readme.md
@@ -597,12 +597,11 @@ Or if you are on [openshift](https://openshift.com):
   <version>${project.version}</version>
 </dependency>
 ```
-
-#### Halkyon Component CRD and  Annotationless configuration
-For Spring Boot applications (i.e. at least one project class is annotated with `@SpringBootApplication`) all you need to do, is adding one of the starters (`io.dekorate:kubernetes-spring-starter` or `io.dekorate:openshift-spring-starter`) to the classpath. No need to specify an additional annotation.
+#### Annotation less configuration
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -615,6 +615,27 @@ The order where these resources are used are:
 
 Here's the full list of supported [configuration options](config.md).
 
+###### Examples
```
cmoulliard: Remarks:
- I'm a bit confused about the location of that documentation example defined with this PR as it seems to appear in the middle of the doc instead to be a well defined section of Halkyon Component CRD for Annotationless
- Instead of writing `Note that some items does not end up in the same hierarchical place where they were configured.`, I would suggest to explain instead : 
  - how a user can use the annotationless mode 
  - how I can override

readme.md
```diff
@@ -597,7 +598,7 @@ Or if you are on [openshift](https://openshift.com):
 </dependency>
 ```
 
-###### Annotation less
+#### Halkyon Component CRD and  Annotationless configuration
 For Spring Boot applications (i.e. at least one project class is annotated with `@SpringBootApplication`) all you need to do, is adding one of the starters (`io.dekorate:kubernetes-spring-starter` or `io.dekorate:openshift-spring-starter`) to the classpath. No need to specify an additional annotation.
```
cmoulliard: This text starting from `For Spring Boot applications ...` has nothing to do with the title `Halkyon Component CRD & Annotationless configuration`

aureamunoz: There is something to do because the annotation less mode is only supported for spring boot applications so far. We use `@SpringBootApplication` to trigger the whole process.

cmoulliard: Then I suggest that you explain/cover that text between the Title `Annotation less` and `Halkyon Component ...` and where you explain what `Annotationless` is/means, how it could be/should be used, How overidding is working, ...

cmoulliard:
Merge this PR and I will review the doc later as this is not enough, you don't really introduce annotationless/explain it, it should be a title level h1 and not h3. We need also for the a Table of Content to more easily navigate within the doc

geoand:
The PR seems like it needs a rebase onto master

readme.md
```diff
@@ -615,6 +615,116 @@ The order where these resources are used are:
...
+  exposeService: false  
+  buildConfig:  
+    type: "docker"  
+    url: "https://github.com/dekorateio/dekorate.git"  
+    ref: "master"  
+    contextPath: "annotations/halkyon-annotations/target/it/"  
+    moduleDirName: "feat-229-override-annotationbased-config"
+
+```
+As in the previous example an additional level `spec` has been introduced for `deploymentMode`.
```
metacosm: We should insist on the fact that the annotation-less configuration completely overrides the annotation configuration as is evidenced by the fact that `exposeService` becomes `false` despite being configured in the annotation… Maybe that needs to be explained above when we say that the annotation-less mode takes precedence, which is note as clear as it should be, imo. 
I think this precedence point will cause issues with our users because I think the natural behavior most people will expect is a merge behavior.

metacosm:
> Merge this PR and I will review the doc later as this is not enough, you don't really introduce annotationless/explain it, it should be a title level h1 and not h3. We need also for the a Table of Content to more easily navigate within the doc

The whole doc should be reviewed, in my opinion as it has grown organically and needs to tending to.

metacosm:
Let's get this merged, I will have more changes to add to it later on.

readme.md
geoand: I guess this answers my question [here](https://github.com/dekorateio/dekorate/issues/313#issuecomment-523367339)

Issue #310, opened 2019-08-16
url of the BuildConfig resource is generated using git+ssh protocol


cmoulliard:
## Issue

The latest version of dekorate (>= 0.7.5) is generating the `BuildConfig URL` using the protocol `git+ssh`. This protocol should be changed to use the default `https` one

**Bad**
```
    buildConfig:
      type: "s2i"
      url: "git+ssh://git@github.com/halkyonio/operator"
```
**Good**
```
    buildConfig:
      type: "s2i"
      url: "https//github.com/halkyonio/operator"
```

aureamunoz:
Hello,
I don't reproduce this with dekorate 0.8-SNAPSHOT, I have the correct url in the generated halkyon files. Could you please give more elements to reproduce it?

cmoulliard:
> Could you please give more elements to reproduce it?

This is because you have cloned the git project of the operator using git+ssh protocol and not https. Change the url within the file `.git/config` and you will be able to reproduce it

cmoulliard:
To reproduce this one, you must use `go get halkyon.io/operator` command

PR #314, opened 2019-08-19
fix: make sure that we use a sanitized version of the git remote URL

metacosm:
Also make sure that the remote URL rewriting works for gitlab as well.

### DIFF
--- a/core/src/test/java/io/dekorate/utils/GitTest.java
+++ b/core/src/test/java/io/dekorate/utils/GitTest.java
@@ -1,74 +1,126 @@
 /**
  * Copyright 2018 The original authors.
- * 
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
- *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


cmoulliard:
LGTM. To be tested of course

aureamunoz:
LGTM. I'm wondering if it would make sense to add a check in the `Issue52Test` in order to validate that `component.getSpec().getBuildConfig().getUrl()` is as expected (I don't know right now if it could break the build during release or in a CI context). WDYT @cmoulliard @metacosm ?

cmoulliard:
> WDYT

This is an excellent idea. You can use a dummy github project from which you know what should be the URL address and then check it @aureamunoz

metacosm:
We could add more tests but there's already unit test coverage.

geoand:
Is suppose this resolves #310?

core/src/test/java/io/dekorate/utils/GitTest.java
```diff
@@ -1,74 +1,126 @@
...
-
-  @Test
-  public void shouldGetUrl() throws Exception {
-    Optional<String> repoUrl = Git.getRemoteUrl(ROOT, Git.ORIGIN);
+  
+  private static Path getRoot(File configFile) {
+    return configFile.toPath().getParent().getParent();
+  }
+  
+  public static void setup(String gitConfig) {
```
geoand: This and the methods below don't have to be `public`. Maybe they should be package-private like the rest?

geoand:
I don't think we need an e2e for this. The feature seems very well covered by unit tests.

Issue #309, opened 2019-08-15
java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 17


drgeb:
Execution failed for task ':compileJava'.
> java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 17

I am not sure what is causing this. Not sure also which files you want to see, let me know I can upload these. Thanks

below is part of the stacktrace

Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 17
        at io.dekorate.project.GradleInfoReader.lambda$readSettingsGradle$1(GradleInfoReader.java:119)
        at io.dekorate.project.GradleInfoReader.readSettingsGradle(GradleInfoReader.java:118)
        at io.dekorate.project.GradleInfoReader.getInfo(GradleInfoReader.java:77)
        at io.dekorate.project.FileProjectFactory.lambda$getProjectInfo$2(FileProjectFactory.java:83)
        at io.dekorate.project.FileProjectFactory.getProjectInfo(FileProjectFactory.java:83)
        at io.dekorate.project.FileProjectFactory.createInternal(FileProjectFactory.java:57)
        at io.dekorate.project.FileProjectFactory.create(FileProjectFactory.java:46)
        at io.dekorate.project.AptProjectFactory.createInternal(AptProjectFactory.java:50)
        at io.dekorate.project.AptProjectFactory.create(AptProjectFactory.java:40)
        at io.dekorate.processor.AbstractAnnotationProcessor.init(AbstractAnnotationProcessor.java:61)
        at org.gradle.api.internal.tasks.compile.processing.DelegatingProcessor.init(DelegatingProcessor.java:57)
        at org.gradle.api.internal.tasks.compile.processing.NonIncrementalProcessor.init(NonIncrementalProcessor.java:33)
        at org.gradle.api.internal.tasks.compile.processing.DelegatingProcessor.init(DelegatingProcessor.java:57)
        at org.gradle.api.internal.tasks.compile.processing.TimeTrackingProcessor.access$301(TimeTrackingProcessor.java:37)
        at org.gradle.api.internal.tasks.compile.processing.TimeTrackingProcessor$4.create(TimeTrackingProcessor.java:88)
        at org.gradle.api.internal.tasks.compile.processing.TimeTrackingProcessor$4.create(TimeTrackingProcessor.java:85)
[...]

PR #326, opened 2019-08-26
fix (#309): readSettingsGradle() now ignores lines that don't contain

iocanel:
the equals sign.


geoand:
What kind of gradle file does this issue fix? If you have an example it would be great to see it :)

iocanel:
Add unit test, that covers the particular use case.

Issue #331, opened 2019-08-28
Configuration found in application configuration should be merged with what defined in annotations


iocanel:
Right now, it completely replaces what specified in annotations. They should be merged instead with priority given to application configuration.

PR #333, opened 2019-08-29
fix (#331) Properly combine annotations with properties.

iocanel:
----
The pull request also fixes some other issues found in the way:

- Handling of service account
- Prevent the same port from being added multiple times.

### DIFF
--- a/core/src/test/java/io/dekorate/utils/BeansTest.java
+++ b/core/src/test/java/io/dekorate/utils/BeansTest.java
@@ -1,0 +1,173 @@
+/**
+ * Copyright 2018 The original authors.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * 
+**/
+
+package io.dekorate.utils;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.assertj.core.util.Arrays;
+import org.junit.jupiter.api.Test;
+
+public class BeansTest {
+
+  private static class Person {
+    private final String name;
+    private final String address;
+    private final String phone;
+
+    public Person() {
+      this(null, null, null);
+    }
+
+    public Person(String name, String address, String phone) {
+      this.name = name;
+      this.address = address;
+      this.phone = phone;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public String getAddress() {
+      return address;
+    }
+
+    public String getPhone() {
+      return phone;
+    }
+  }
+
+  private static class Team {
+    private final String name;
+    private final Person captain;
+    private final List<Person> members;
+
+    public Team() {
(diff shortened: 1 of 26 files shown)


geoand:
Do you feel we have enough tests for this feature?

iocanel:
> Do you feel we have enough tests for this feature?

Well, with the existing tests we actually test that:

- The order suppliers are processed is correct.
- Merging of annotations and properties don't break existing annotation-less tests.
- Bean merging core rules are respected.

To be honest, I can't think of another case we need to cover at this point.
We will add more regression tests as we find issues.

core/src/test/java/io/dekorate/utils/BeansTest.java
```diff
@@ -0,0 +1,145 @@
...
+    Team team2 = new Team("A", null, members);
+    Team combined = Beans.combine(team1, team2);
+    assertNotNull(combined);
+    assertEquals("A", combined.getName());
+    assertNotNull(combined.getMembers());
+    assertEquals(member1, combined.getMembers().get(0));
+  }
+
+  @Test
+  public void shouldCombineList() throws Exception {
```
geoand: maybe a test for maps as well?

iocanel: Here's a good one:

It's not even possible to have Sets and Maps inside those Objects, given that they are generated from annotations (atm). 

So, I guess we either have to remove handling these cases in the code or test them.

Let me do the later as its safer.

geoand: :+1:

iocanel:
@geoand: PR updated.

geoand:
If CI is happy, I'm happy :)

I think @metacosm and @aureamunoz should take a closer look since they have more experience than I do at this point :P

core/src/test/java/io/dekorate/utils/BeansTest.java
```diff
@@ -0,0 +1,172 @@
...
+    Person captain = new Person("Jim", null, null);
+    Person member1 = new Person("Joe", null, null);
+    Team team1 = new Team("A", captain, null);
+    List<Person> members = new ArrayList<>();
+    members.add(member1);
+    Team team2 = new Team("A", null, members);
+    Team combined = Beans.combine(team1, team2);
+    assertNotNull(combined);
+    assertEquals("A", combined.getName());
+    assertNotNull(combined.getMembers());
```
aureamunoz: This is really minor/secondary but I would add a check about the captain of combined because it helps to validate for my (and maybe is the same for others) assumptions about the code : `assertEquals(captain, combined.getCaptain());`

frameworks/spring-boot/src/it/issue-331/src/test/java/io/dekorate/annotationless/Issue331Test.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class Issue331Test {
+
+  @Test
+  public void shouldContainDeployment() {
```
aureamunoz: shouldContainDeploymentWithCombinedValues?

frameworks/spring-boot/src/it/issue-331/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>2.1.5.RELEASE</version>
+    <relativePath/> <!-- lookup parent from repository -->
+  </parent>
+  <groupId>io.dekorate</groupId>
+  <artifactId>issue-331</artifactId>
```
aureamunoz: I propose a more self-explained name, for example: issue-331-annotations-combined-with-properties

iocanel:
Since all feedback had been applied, I will merge this.

PR #336, opened 2019-09-04
Thorntail

Ladicek:

PR #341, opened 2019-09-18
Thorntail pt2.

iocanel:
Github farted and we can't reopen this one: https://github.com/dekorateio/dekorate/pull/336 so I created a new one.

### DIFF
--- a/annotations/option-annotations/src/main/java/io/dekorate/option/configurator/ApplyJvmOptsConfigurator.java
+++ b/annotations/option-annotations/src/main/java/io/dekorate/option/configurator/ApplyJvmOptsConfigurator.java
@@ -45,6 +45,7 @@ import java.util.Optional;
 public class ApplyJvmOptsConfigurator extends Configurator<BaseConfigFluent<?>> {
 
   private static final String JAVA_OPTS = "JAVA_OPTS";
+  private static final String JAVA_OPTIONS = "JAVA_OPTIONS";
 
   private static final String EQ = "=";
   private static final String EMPTY = " ";
@@ -73,17 +74,24 @@ public class ApplyJvmOptsConfigurator extends Configurator<BaseConfigFluent<?>> 
   public void visit(BaseConfigFluent<?> kubernetesConfig) {
     JvmConfig config = this.config.get();
 
-    Optional<String> existing = Arrays.stream(kubernetesConfig.getEnvVars()).filter(e -> e.getName().equals(JAVA_OPTS)).map(Env::getValue)
+    setJavaOptsEnvVar(JAVA_OPTS, kubernetesConfig, config);
+    setJavaOptsEnvVar(JAVA_OPTIONS, kubernetesConfig, config);
+  }
+
+  private void setJavaOptsEnvVar(String envVar, BaseConfigFluent<?> kubernetesConfig, JvmConfig jvmConfig) {
+    Optional<String> existing = Arrays.stream(kubernetesConfig.getEnvVars())
+      .filter(e -> e.getName().equals(envVar))
+      .map(Env::getValue)
       .findFirst();
 
     if (existing.isPresent()) {
-      kubernetesConfig.editMatchingEnvVar(e -> e.getName().equals(JAVA_OPTS))
-        .withValue(mergeOptions(existing.get(), config))
+      kubernetesConfig.editMatchingEnvVar(e -> e.getName().equals(envVar))
+        .withValue(mergeOptions(existing.get(), jvmConfig))
         .endEnvVar();
     } else {
       kubernetesConfig.addNewEnvVar()
-        .withName(JAVA_OPTS)
-        .withValue(mergeOptions("", config))
+        .withName(envVar)
+        .withValue(mergeOptions("", jvmConfig))
         .endEnvVar();
     }
   }
--- a/design.md
+++ b/design.md
@@ -53,7 +53,7 @@ The base interface that all `model` objects implement is the HasMetadata. For ex
 An object / pojo that encapsulates the information provided by an annotation and the project (e.g. name, version etc).
 An example configuration is the `KubernetesConfig` class:
 
-    public class KubernetecConfig  {
+    public class KubernetesConfig  {
         private String group;
         private String name;
         private String version;
@@ -79,7 +79,7 @@ For example a configurator that can be used to add a label to `KubernetesConfig`
 | Configurator           | Target              | Description                                                                   |
 |------------------------|---------------------|-------------------------------------------------------------------------------|
 | ApplyOpenshiftConfig   | SourceToImageConfig | Applies group, name and version from OpenshiftConfig to SourceToImage config. |
-| AddPort                | KubernetesConfig    | adds a port to all containers.                                                |
+| AddPort                | KubernetesConfig    | Adds a port to all containers.                                                |
 | ApplyDockerBuildHook   | DockerBuildConfig   | Apply the docker build hook configuration.                                    |
 | ApplySourceToImageHook | SourceToImageConfig | Apply source to image build hook.                                             |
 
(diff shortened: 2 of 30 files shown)


iocanel:
@Ladicek Not sure what's going on with the old PR. I can't reopen it and TBH I don't even recall willingly closing it in the first place.

If CI passes I'll merge ASAP before we have more surprises.

Ladicek:
OK, thanks!

iocanel:
@Ladicek: I gave a glimpse at the test failures. It seems that the readiness check which is a pre-req for the test to start times out. 

So, I assume that the generated deployment lacks a readiness probe. I would add a probe manually using the annotations. This is what we do at: 
https://github.com/dekorateio/dekorate/blob/master/examples/spring-boot-on-kubernetes-example/src/main/java/io/dekorate/example/sbonkubernetes/Main.java#L24

You can add it to the example, or I can add it to your PR myself whatever you prefer.

Ladicek:
@iocanel You mean like this https://github.com/dekorateio/dekorate/pull/341/files#diff-121d020b7ed8d79cc944b3136e9c2016R28 and this https://github.com/dekorateio/dekorate/pull/341/files#diff-2bc2ed91c5e55e246c76ac5e79d07127R28 ? :-) As I said, the tests are passing for me locally, using Minishift. In CI, "Thorntail on Kubernetes" passes as well, it's "Thorntail on Openshift" that fails. I'm not sure what could possibly timeout there. Is there a way to interactively debug the test in Circle CI?

iocanel:
So, the probes are already set! Yeah, they seem right.

Let me look at the logs again.

iocanel:
At some point the Xmx value looked suspicious, but it does work  for other tests, so its not.

I'll have to checkout the PR and check closer. Maybe, we need to enhance our testing stuff so that the provide richer feedback.

Ladicek:
Hmm, it _might_ be wrong, because it's actually `-Xms`. I copied it from https://github.com/dekorateio/dekorate/blob/master/examples/spring-boot-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java#L26 without thinking, but it might have been wrong there all the time...

iocanel:
@Ladicek can you please rebase?

Ladicek:
Rebased.

iocanel:
Thanks! Hopefully, we will now get more information about the nature of the test failures.

Ladicek:
Thanks!

Ladicek:
The test still fails and I don't see any meaningful info in the log. I will start adding debug logs all around the place in temporary commits, if you guys are fine with that :-)

iocanel:
@Ladicek: It seems that the pod is getting restarted, despite the fact that the log seems normal. Maybe, the liveness/readiness enpoint is not available?

Ladicek:
Sounds strange. It keeps working just fine locally. And the pod log _doesn't_ look normal, there's only half of what should be there.

Ladicek:
Hmm, now that is interesting:

```
2019-09-26 12:27:34,902 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-4) MSC000001: Failed to start service org.wildfly.undertow.listener.default: org.jboss.msc.service.StartException in service org.wildfly.undertow.listener.default: WFLYUT0082: Could not start 'default' listener.
	at org.wildfly.extension.undertow.ListenerService.start(ListenerService.java:211)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:1738)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1700)
	at org.jboss.msc.service.ServiceControllerImpl$ControllerTask.run(ServiceControllerImpl.java:1558)
	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:1985)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1487)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1364)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.SocketException: Protocol family unavailable
	at sun.nio.ch.Net.bind0(Native Method)
	at sun.nio.ch.Net.bind(Net.java:433)
	at sun.nio.ch.Net.bind(Net.java:425)
	at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
	at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
	at org.xnio.nio.NioXnioWorker.createTcpConnectionServer(NioXnioWorker.java:178)
	at org.xnio.XnioWorker.createStreamConnectionServer(XnioWorker.java:303)
	at org.wildfly.extension.undertow.HttpListenerService.startListening(HttpListenerService.java:106)
	at org.wildfly.extension.undertow.ListenerService.start(ListenerService.java:199)
	... 8 more
```

iocanel:
@Ladicek: Then maybe we need to add some `@JavaOptions` to set the prefered stack to ipv4?

Ladicek:
This typically occurs because IPv6 and the canonical solution is to enforce IPv4, but we already do that: `@JvmOptions(server = true, xms = 1024, preferIPv4Stack = true, secureRandom = SecureRandomSource.NonBlocking)`. It's also in the YAML:

```yaml
- apiVersion: "apps.openshift.io/v1"
  kind: "DeploymentConfig"
  metadata:
    ...
  spec:
    ...
    template:
      metadata:
        ...
      spec:
        containers:
        - env:
          - name: "KUBERNETES_NAMESPACE"
            valueFrom:
              fieldRef:
                fieldPath: "metadata.namespace"
          - name: "JAVA_APP_JAR"
            value: "/deployments/thorntail-on-openshift-example-0.8-SNAPSHOT-thorntail.jar"
          - name: "AB_PROMETHEUS_OFF"
            value: "true"
          - name: "JAVA_OPTS"
            value: "-Xms=1024M -Djava.security.egd=file:/dev/urandom -server -Djava.net.preferIPv4Stack=true"
        ...
```

Ladicek:
It's also in the debug log (newlines mine):

```
env=[
EnvVar(name=KUBERNETES_NAMESPACE, value=null, valueFrom=EnvVarSource(configMapKeyRef=null, fieldRef=ObjectFieldSelector(apiVersion=v1, fieldPath=metadata.namespace, additionalProperties={}), resourceFieldRef=null, secretKeyRef=null, additionalProperties={}), additionalProperties={}),
EnvVar(name=JAVA_APP_JAR, value=/deployments/thorntail-on-openshift-example-0.8-SNAPSHOT-thorntail.jar, valueFrom=null, additionalProperties={}),
EnvVar(name=AB_PROMETHEUS_OFF, value=true, valueFrom=null, additionalProperties={})
EnvVar(name=JAVA_OPTS, value=-Xms=1024M -Djava.security.egd=file:/dev/urandom -server -Djava.net.preferIPv4Stack=true, valueFrom=null, additionalProperties={})
]
```

Ladicek:
Perhaps the S2I image we use doesn't pick up the `JAVA_OPTS` env var? I'm personally not a fan of the Fabric8 images, but there's nothing better I guess. I'll investigate.

Ladicek:
Indeed the Java S2I image doesn't expect `JAVA_OPTS`; it expects `JAVA_OPTIONS`.

Ladicek:
I was able to confirm locally that changing to `JAVA_OPTIONS` makes the `run-java` crap pick up whatever was configured in `@JvmOptions`. Pushed a temporary commit to verify with CI.

Ladicek:
# DO NOT MERGE  YET

Yay, it passed now! This confirms that the S2I image (the Fabric8 one, but I know the Red Hat ones are identical in this regard) doesn't expect the `JAVA_OPTS` env var; instead, it expects `JAVA_OPTIONS`.

@iocanel Now the question is, what do we want to do about it? I know `JAVA_OPTS` is more of a "standard" elsewhere, but the S2I images are different. We can easily move from `JAVA_OPTS` to `JAVA_OPTIONS` (that's what I did in the last commit), but that would technically be a breaking change. WDYT?

iocanel:
@Ladicek: I think that the easiest and lowest risk solution is to set both `JAVA_OPTS` and `JAVA_OPTIONS`.

Ladicek:
That sounds a little strange, but indeed would probably be safest. I can't think of a case when somehow both env vars would be consumed. Either there's the Fabric8 Java image, which consumes `JAVA_OPTIONS` and directly runs a JAR (which almost certainly doesn't consume `JAVA_OPTS`), or there's some other custom image, which most likely doesn't know about `JAVA_OPTIONS` at all.

Will redo the last commit.

iocanel:
And even if there is an app that for any reason uses both, the worst thing that can happen is to end up passing dublicate options to java (which btw doesn't seem to cause any issue).

Not a solution that we will brag about, but definitely the safest.

Ladicek:
Done, and ready for review.

iocanel:
Awesome work @Ladicek!
Thank you very much!

Ladicek:
Thanks! :-)

PR #342, opened 2019-09-20
Decoupling platform from build method.

iocanel:
----

The idea is that the build method should not be coupled to platform.

Some examples:

- I should be able to do docker builds on Openshift.
- I should be able to do docker and s2i builds on Knative.
- I should be able to do jib builds on all platforms.


geoand:
Neat idea!

Are there any examples where we can see it in action?

iocanel:
> Neat idea!
> 
> Are there any examples where we can see it in action?

Not fully demonstratable yet.

I'd like to merge it as is though, cause it does bring with it some integration test improvements which are needed by other pull requests.

geoand:
> > Neat idea!
> > Are there any examples where we can see it in action?
> 
> Not fully demonstratable yet.
> 
> I'd like to merge it as is though, cause it does bring with it some integration test improvements which are needed by other pull requests.

You have my vote!

PR #344, opened 2019-09-26
Decouple builds

iocanel:
In this iteration we finally split docker and s2i build services into different modules.
We also clean up and move some code around to align with the split.
Finally, we do some improvements in the S2iBuildService to make sure it works correctly in all cases.


iocanel:
This pull request has grown way bigger than initally inteded.

The original goal was to "**decouple docker and s2i builds from kubernetes and openshift respectively**". However, this change did brought to surface a large number of issues mostly related to annotationless, so I had to include changes like "**change the way generators are loaded**". Also we did have quite a few failures in the CI land as a result of updating the kubernetes-client, so I also had to "**workaround client issues inside our junit extensions**".

At least, the commits are relatively small and possibly make sense if examined one by one.

geoand:
In my experience, with monster PRs like these, there are only 2 ways to go:

1) reject
2) approve without much fuss

Given the source of this PR, I say the latter is in order :)

Issue #369, opened 2019-10-14
@OpenshiftApplication livenessProbe/readinessProbe always uses default parameters


liborfuka:
It always uses default parameters even i set 
@OpenshiftApplication(expose=true, 
                         livenessProbe=@Probe(httpActionPath="/actuator/health",initialDelaySeconds=60,periodSeconds=3,timeoutSeconds=1),
                      readinessProbe=@Probe(httpActionPath="/actuator/health",initialDelaySeconds=10,periodSeconds=10,timeoutSeconds=1)).

More investigation: when i don't have spring-boot-actuator dependency in pom.xml it sets correct parameters. When i add spring-boot-actuator dependency to pom.xml it always uses defaults 0,30,10.

I used dekorate-core-0.8.2 for testing.

iocanel:
Great catch! This is quite serious.

It seems that we have a conflict between two decorators, one added by the `OpenshiftHanlder` that handle the annotation and one added by the `SpringBootApplicationHandler`. 

The issue here is that in both cases a configurator should be used instead and the configuration provided by each should be merged.

PR #373, opened 2019-10-14
fix (#369): User probe configuration is now used even when spring

iocanel:
actuator is available.


frameworks/spring-boot/src/it/issue-370-probe-properties/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM openjdk:8u171-alpine3.7
```
geoand: Doesn't the PR fix issue 369, not issue 370? Asking because the directory is named `issue-370-probe-properties`

PR #377, opened 2019-10-16
refactor: Remove image related properties from application

iocanel:
configuration configuration and vice versa.


geoand:
@iocanel can you explain this one a little? I feel like I am missing some key piece

iocanel:
@geoand: Sure let me try.

We have separated application configuration from image configuration. The main reason behind this choice is to be able to mix and match.  This means that annotations like `@KubernetesApplication` no longer has properties like `autoBuildEnabled` or `dockerfile` etc and also annotations like `@DockerBuild` no longer have `autoDeployEnabled`.

This is the goal of this pull request. The rest of the code changes were added to support this. For example the junit extensions where reading everything from the application config.

annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/listener/KubernetesSessionListener.java
```diff
@@ -50,20 +50,21 @@
...
 
-    KubernetesConfig kubernetesConfig = config.get();
+    KubernetesConfig kubernetesConfig = optionalAppConfig.get();
     Resources resources = session.resources();
     KubernetesList generated = session.getGeneratedResources().getOrDefault(KUBERNETES, new KubernetesList());
 
     BuildService buildService = null;
-    if (kubernetesConfig.isAutoPushEnabled() || kubernetesConfig.isAutoBuildEnabled() || kubernetesConfig.isAutoDeployEnabled()) {
+    ImageConfiguration imageConfig = optionalImageConfig.orElse(null);
+    if (imageConfig.isAutoPushEnabled() || imageConfig.isAutoBuildEnabled() || kubernetesConfig.isAutoDeployEnabled()) {
```
geoand: Can `imageConfig` be null here?

geoand: Actually it looks like it can't so maybe just use `optionalImageConfig.get()`?

examples/spring-boot-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java
```diff
@@ -15,15 +15,11 @@
...
 
-import io.dekorate.kubernetes.annotation.Env;
 import io.dekorate.openshift.annotation.OpenshiftApplication;
-import io.dekorate.option.annotation.JvmOptions;
-import io.dekorate.option.annotation.SecureRandomSource;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@OpenshiftApplication(buildEnvVars = @Env(name="foo", value = "bar"))
-@JvmOptions(server = true, preferIPv4Stack = true, secureRandom = SecureRandomSource.NonBlocking)
```
geoand: Was this example suppose to show the use of these? If not, does it still make sense to keep it?

examples/spring-boot-with-svcat-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java
```diff
@@ -15,15 +15,14 @@
...
-import io.dekorate.kubernetes.annotation.Env;
 import io.dekorate.openshift.annotation.OpenshiftApplication;
 import io.dekorate.servicecatalog.annotation.ServiceCatalog;
 import io.dekorate.servicecatalog.annotation.ServiceCatalogInstance;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@OpenshiftApplication(buildEnvVars = @Env(name="foo", value = "bar"))
+@OpenshiftApplication
```
geoand: Does this example still make sense without the build env vars?

geoand:
@iocanel Thanks for the explanation!

I added some minor comments. Also CI seems to not be happy

examples/spring-boot-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java
iocanel: IMHO, our examples are bloated with stuff that should be part of integration tests instead, so these need to go.

examples/spring-boot-with-svcat-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java
iocanel: Same as above

annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/listener/KubernetesSessionListener.java
iocanel: Agree.

examples/spring-boot-on-openshift-example/src/main/java/io/dekorate/example/sbonopenshift/Main.java
geoand: Gotcha, thanks

geoand:
Weird... Github UI shows that there are conflicts now...

iocanel:
Yeah, I need to rebase, but also I need to fix that tests, so it will take a while.

geoand:
kk

PR #386, opened 2019-10-25
Stop defaulting to /actuator/health on liveness probes

Albertoimpl:
At the moment, every example using Spring boot defaults to `livenessProbe` == `readinessProbe` == `/actuator/health` and that is typically a bad idea.
Liveness Probes responds to the question of when should I restart a container. Readiness Probes responds to the question of when should I start accepting traffic.

Making Liveness dependent on an external check, like other services or databases can cause a restart of all your containers when just a single external failure occurs.

For that reason, I suggest making the liveness probe defaults to `/actuator/info`
What are your thoughts?

Thanks,
Alberto


iocanel:
@Albertoimpl thank you for this. 
I like your suggestion. Let's see what others think about it.

@geoand? @ryanjbaxter?

geoand:
Makes sense

Albertoimpl:
Thanks, @iocanel and @geoand!

PR #393, opened 2019-10-30
 Serialization options

iocanel:
Genreated resources are no longer wrapped in a KubernetesList.

Requires a new release of the kubernetes-client and thus marked as WIP.


geoand:
Very nice usability improvement!

PR #407, opened 2019-11-22
[WIP] feat: Integrate with fmp out of the box.

iocanel:

PR #426, opened 2019-12-05
feat: Source/input manifests are supported

manusa:
Improvement to allow the use of source manifests within Dekorate.

Source manifests are deserialized and later on merged with manifests generated by Dekorate handlers.

Legacy versions of Fabric8 Maven Plugin generated manifests are automagically detected (zero-config) and merged. Manifest output directory is also switched to FMP's output directory if FMP has been detected.


core/src/main/java/io/dekorate/SessionReader.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.dekorate;
+
+public interface SessionReader {
+
+  void read(Session session, Resources resources);
```
iocanel: There is no need to pass both Session and Resources, as one can grab the resources from the Session. See `SessionWritter`.

core/src/main/java/io/dekorate/Session.java
```diff
@@ -202,12 +212,17 @@ public void addListener(SessionListener listener) {
       LOGGER.info("Generating manifests.");
       closed.set(true);
       populateFallbackConfig();
+      populateInitialConfig();
```
iocanel: I'd rather pickup the extenral configuration, during Session initalization.
To be fair, it won't make much of a difference (as it will still run before all handlers), still it seems somehow cleaner.

Also, it should definitely run before adding fallback config.

annotations/knative-annotations/src/main/java/io/dekorate/knative/generator/KnativeApplicationGenerator.java
```diff
@@ -75,6 +76,7 @@ default void on(ConfigurationSupplier<KnativeConfig> config) {
       Session session = getSession();
       session.configurators().add(config);
       session.handlers().add(new KnativeHandler(session.resources(), session.configurators()));
+      session.resources().groups().putIfAbsent(KNATIVE, new KubernetesListBuilder());
```
iocanel: Why?

I'd rather have generators not directly mess with the resources if possible.

examples/spring-boot-with-source-manifests-example/src/main/java/io/dekorate/examples/sbsourcemanifests/Main.java
```diff
@@ -0,0 +1,34 @@
...
+package io.dekorate.examples.sbsourcemanifests;
+
+import io.dekorate.option.annotation.GeneratorOptions;
+import io.dekorate.prometheus.annotation.EnableServiceMonitor;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.web.bind.annotation.RestController;
+
+@GeneratorOptions(inputPath = "source-manifests")
+@EnableServiceMonitor
```
iocanel: With a quick glimpse this annotation doesn't have a purpose in this example, so I'd rather have it removed.

examples/spring-boot-with-source-manifests-example/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <java.version>1.8</java.version>
+
+    <version.spring-boot>2.1.6.RELEASE</version.spring-boot>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.dekorate</groupId>
+      <artifactId>openshift-spring-starter</artifactId>
```
iocanel: If we there is not particular reason to use `openshift` in this example I'd rather use kuberentes. Also, I'd like to have the `on-kubernetes` as part of the folder name and artifactId.

examples/spring-boot-with-source-manifests-example/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>examples</artifactId>
+    <groupId>io.dekorate</groupId>
+    <version>0.10-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>io.dekorate</groupId>
+  <artifactId>spring-boot-with-source-manifests-example</artifactId>
```
iocanel: Can we use `exisitng` instead of `source`?

annotations/knative-annotations/src/main/java/io/dekorate/knative/generator/KnativeApplicationGenerator.java
manusa: TBH this was ported from your [PR](https://github.com/dekorateio/dekorate/pull/407/files) and have no idea of what it's actually doing.
Same for: 
- https://github.com/dekorateio/dekorate/pull/426/files#diff-fb310640781383297c9efb65ecea11a6
- https://github.com/dekorateio/dekorate/pull/426/files#diff-e69a70d2513b994051c752bfba6f0125

Please, confirm if I should remove those too.

core/src/main/java/io/dekorate/Session.java
manusa: After checking alternatives, existing manifests must be read in this method so that AnnotationProcessors in charge of setting input and output directories have already been triggered. i.e. It's not possible to read these existing manifests during session initialization as input directory will still not be declared.

I will move it before `populateFeedbackConfig` and rename it to `readExistingResources` (previous name was devised from a former iteration and doesn't make much sense)

annotations/knative-annotations/src/main/java/io/dekorate/knative/generator/KnativeApplicationGenerator.java
iocanel: Oooops! I really hate it when others do that to me!
Sorry, about that!

iocanel: keep them for now

Issue #420, opened 2019-12-02
Specify external registry


jponge:
It would be useful to have the option to specify a registry.

I am trying to reproduce this with Dekorate, but this is not possible:

```yaml
spec:
  containers:
    - image: localhost:32000/rain-service
```

iocanel:
You should be able to specify a registry with something like:

    dekorate:
      docker:
        registry: localhost:32000

jponge:
Given `application.yml`:

```yaml
dekorate:
  kubernetes:
    serviceType: LoadBalancer
    ports:
      - name: HTTP
        containerPort: 8080
    liveness-probe:
      http-action-path: /health
      initial-delay-seconds: 3
      period-seconds: 5
  docker:
    registry: localhost:32000

```

I do not get any mention of the registry:

```yaml
---
apiVersion: "v1"
kind: "Service"
metadata:
  labels:
    app: "rain-service"
    version: "0-SNAPSHOT"
    group: "jponge"
  name: "rain-service"
spec:
  ports:
  - name: "HTTP"
    port: 8080
    targetPort: 8080
  selector:
    app: "rain-service"
    version: "0-SNAPSHOT"
    group: "jponge"
  type: "LoadBalancer"
---
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  labels:
    app: "rain-service"
    version: "0-SNAPSHOT"
    group: "jponge"
  name: "rain-service"
spec:
  replicas: 1
  selector:
    matchLabels:
      app: "rain-service"
      version: "0-SNAPSHOT"
      group: "jponge"
  template:
    metadata:
      labels:
        app: "rain-service"
        version: "0-SNAPSHOT"
        group: "jponge"
    spec:
      containers:
      - env:
        - name: "KUBERNETES_NAMESPACE"
          valueFrom:
            fieldRef:
              fieldPath: "metadata.namespace"
        image: "jponge/rain-service:0-SNAPSHOT"
        imagePullPolicy: "IfNotPresent"
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: "/health"
            port: 8080
            scheme: "HTTP"
          initialDelaySeconds: 3
          periodSeconds: 5
          successThreshold: 1
          timeoutSeconds: 10
        name: "rain-service"
        ports:
        - containerPort: 8080
          name: "HTTP"
          protocol: "TCP"
```

iocanel:
@jponge: can you please let me know which version you are using?

[...]

PR #436, opened 2020-01-14
fix (#420): Explicit configuration is always considered applicable.

iocanel:
We recently added support for multple different image configurations (docker, jib, s2i).
Each config is meant to be used only in projects where its actually applicable.

For instance it doesn't make sense to use docker config, if no docker files are provided.

However, if the use has provided explicitly configuration for cases like this, we should consider it applicable no matter what.

### DIFF
--- a/annotations/docker-annotations/src/main/java/io/dekorate/docker/buildservice/DockerBuildServiceFactory.java
+++ b/annotations/docker-annotations/src/main/java/io/dekorate/docker/buildservice/DockerBuildServiceFactory.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  * 
-**/
+ **/
 
 package io.dekorate.docker.buildservice;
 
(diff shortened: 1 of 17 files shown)


annotations/docker-annotations/src/main/java/io/dekorate/docker/buildservice/DockerBuildServiceFactory.java
```diff
@@ -62,4 +65,13 @@ public BuildServiceApplicablility checkApplicablility(Project project, ImageConf
         : String.format(MESSAGE_NOK, project.getRoot().resolve(dockerFile));
 		return new BuildServiceApplicablility(applicable, message);
 	}
+
+	@Override
+	public BuildServiceApplicablility checkApplicablility(Project project, ConfigurationSupplier<ImageConfiguration> supplier) {
+    if (supplier.isExplicit()) {
```
geoand: The indentation seems a little off here

annotations/docker-annotations/src/main/java/io/dekorate/docker/buildservice/DockerBuildServiceFactory.java
```diff
@@ -62,4 +65,13 @@ public BuildServiceApplicablility checkApplicablility(Project project, ImageConf
...
+      return new BuildServiceApplicablility(true, MESSAGE_OK);
+    }
+
+    return checkApplicablility(project, supplier.get());
```
geoand: Same here

geoand:
CI also seems to have failed

iocanel:
@geoand both issues have been addressed.

PR #441, opened 2020-01-21
fix: Images.getTag returns meaningful value when tag is missing

Ladicek:


core/src/main/java/io/dekorate/utils/Images.java
```diff
@@ -103,12 +103,24 @@ public static String getName(String image) {
...
    * Return the tag of the image.
    * @param image The docker image.
-   * @return The tag if present or null otherwise.
+   * @return The tag if present, {@code latest} otherwise.
    */
   public static String getTag(String image) {
     if (image.contains(COLN)) {
       return image.substring(image.indexOf(COLN) + 1);
     }
+    return "latest";
```
Ladicek: Alternatively, we could make it return `null` as the javadoc used to say, but then all consumers (of which there's 1, AFAIK) would have to handle that. I think `latest` is the customary default value.

Ladicek:
My motivation is that in a testing Quarkus application, I set `s2i.builder-image=registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift` (without the tag) and got this in `target/kubernetes/openshift.yml`:

```yaml
apiVersion: "build.openshift.io/v1"
kind: "BuildConfig"
...
spec:
  ...
  strategy:
    sourceStrategy:
      from:
        kind: "ImageStreamTag"
        name: "openjdk18-openshift:registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift"
```

Issue #442, opened 2020-01-31
Address inconsistencies with handing of group, name, version and labels.


iocanel:
At the moment we don't have a clear strategy on how to handle group, name and version and that is causing confusion.

In same cases when we define on of `group`, `name`, `version`, and `labels` from one platform (e.g `kubernetes` these are automatically applied to others too). This is not necessarily bad but it contrasts with how we handle the rest of the properties.

In other cases it allows resources to become out of sync.

Some issues caused by that:

- https://github.com/quarkusio/quarkus/issues/6726
- https://issues.redhat.com/projects/ENTSBT/issues/ENTSBT-205

What we need to do is to bound this properties per platform.
This does have a small side effect on how to register some of our `decorators` especially those that are `named`. 

We should make named decorators smarter and use the metadata of the resource on which they are applied.

PR #443, opened 2020-01-31
fix (#442): Bound name, group, version and labels per platform and

iocanel:
don't have global values for them.


geoand:
Shouldn't we have new tests for this?

iocanel:
> Shouldn't we have new tests for this?

geoand:
Ok cool 👍

iocanel:
> LGTM, I am only requesting changes so we can get the tests and not merge by accident without them

Tests have been added!

PR #448, opened 2020-02-04
chore: Make Project, BuildInfo and ScmInfo immutable and remove uneed…

iocanel:
…ed methods.


aureamunoz:
What about the CI problem? There is a problem with gradle on Openshift,

PR #449, opened 2020-02-05
fix (#442): S2iBuildService now uses whatever name was given to the

iocanel:
resource. S2i resources aligned with the rest of the converntions. Added tests.

This pull request is meant to do the following:

- Have openshift resources correctly refer to s2i resources  (imagestreams, etc).
- Have openshift shutdown hooks, correctly identify s2i names.
- Have s2i build service retain imagestreams (workaround ci related issues).
- Add integration test for the above.
- Deal with circleci issues and missing s2i-java tags.

### DIFF
--- a/tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameIT.java
+++ b/tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameIT.java
@@ -6,7 +6,7 @@ import io.dekorate.testing.annotation.Inject;
 import io.dekorate.testing.openshift.annotation.OpenshiftIntegrationTest;
 
 @OpenshiftIntegrationTest
-public class Issue442BuildConfigNameIntegrationTest {
+public class Issue442BuildConfigNameIT {
 
   @Test
   void shouldBeUpAndRunning() {
--- a/tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameTest.java
+++ b/tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameTest.java
@@ -30,12 +30,12 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 public class Issue442BuildConfigNameTest {
 
   @Test
-  public void shouldHaveCustomBuildConfigNameYaml() {
+  public void shouldHaveModuleBuildConfigNameYaml() {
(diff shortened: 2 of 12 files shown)


tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameIT.java
```diff
@@ -0,0 +1,15 @@
+package io.dekorate.examples.kuberentes;
```
geoand: It seems like the package is misspelled. Not really important of course, just noting it so you can take care of it if you like

tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameTest.java
```diff
@@ -35,7 +35,7 @@ public void shouldHaveCustomBuildConfigNameYaml() {
     assertNotNull(list);
     BuildConfig b = findFirst(list, BuildConfig.class).orElseThrow(() -> new IllegalStateException());
     assertNotNull(b);
-    assertEquals("o-name", b.getMetadata().getName());
+    assertEquals("issue-442-build-config-name", b.getMetadata().getName());
```
geoand: I'm not really sure I understand where this is coming from, can you please elaborate?

tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameIT.java
iocanel: Fixed.

geoand: Thanks

tests/issue-442-build-config-name/src/test/java/io/dekorate/examples/kuberentes/Issue442BuildConfigNameTest.java
iocanel: When we use a custom name for Openshift resources it should only affect things like: 
- DeploymentConfig
- Service
- Route

It should **NOT** affect BuildConfig and ImageStreams. Why?
Because all things s2i are customized independently.

This is what this test case is about.

geoand: Great, thanks for the detailed explanation

PR #453, opened 2020-02-15
Utility improvements for downstream projects

iocanel:
This pull request adds/changes some utility that are really handy in downstream projects (e.g. quarkus) but doesn't make sense to host there.


iocanel:
@geoand: "I am calling it a day? I lied!"

Issue #409, opened 2019-11-24
Can't manipulate container image property in generated Kubernetes yaml file


tguless:
I need my generated yaml file to reference a different image depending on which spring profile I choose. Specifically, I want it to point to AWS ECR when using one profile, versus my local repo in other cases. 

I am unable to do this - I can't seem to granularly control the container image attribute using application.yml file. 

Setting group proprerty was setting the image name prefix initially, but I believe after upgrading to 0.9.9 even that stopped working. 

Does anyone have any input on how I can set the image name from application.yml properties file?

On my mac it keeps going back to using my login id:

         image: "gule001/sme-mc-qm-svc:1.0"

iocanel:
@tguless: AFAIR, the issue with group not working in `0.9.9` has been fix on master and will be part of the next release   (possibly today).

The are the following properites you can use:
- `dekorate.kubernetes.group`
- `dekorate.kubernetes.name`
- `dekorate.kubernetes.version`
- `dekorate.docker`

Now, regardling profiles.... Dekorate generates manifests on compile time, while you choose the profile of your spring profile on runtime (after the manifests) have been generated.

Also, `dekorate` will not read at all profile specific configuration and will just read `application.properties`, `application.yml` and `application-kubernetes.properties/yml`.

iocanel:
We need an integration test for this use case.

iocanel:
Added a test and it seems that if fails!
So, contrary to my belief this is still an issue.

sasiperi:
> @tguless: AFAIR, the issue with group not working in `0.9.9` has been fix on master and will be part of the next release (possibly today).
> 
> The are the following properites you can use:
> 
> * `dekorate.kubernetes.group`
> * `dekorate.kubernetes.name`
> * `dekorate.kubernetes.version`
> * `dekorate.docker`
> 
> Now, regardling profiles.... Dekorate generates manifests on compile time, while you choose the profile of your spring profile on runtime (after the manifests) have been generated.
> 
> Also, `dekorate` will not read at all profile specific configuration and will just read `application.properties`, `application.yml` and `application-kubernetes.properties/yml`.

This did not work for me for "group" override, after looking at the source code I realized the prop name used is not "gorup" for kubes but is "partOf", for it to work.
So what worked for me is the following property, to override "group" of the image for kubernetes annotations

The name that worked in app.props (so the root/image-name or on mac the "userid/image-name) is
[...]

PR #457, opened 2020-02-27
Add dekorate.docker.group property so making image manipulation clea…

jbeaken:
…rer, see https://github.com/dekorateio/dekorate/issues/409

### DIFF
--- a/examples/spring-boot-on-kubernetes-without-annotations/src/main/resources/application.properties
+++ b/examples/spring-boot-on-kubernetes-without-annotations/src/main/resources/application.properties
@@ -8,3 +8,4 @@ dekorate.kubernetes.ports[0].containerPort=8080
 dekorate.kubernetes.serviceType=LoadBalancer
 dekorate.kubernetes.configMapVolumes[0].volumeName=bar-volume
 dekorate.kubernetes.configMapVolumes[0].configMapName=foo-map
+dekorate.docker.group=dekorate.io


examples/spring-boot-on-kubernetes-without-annotations/src/main/resources/application.properties
```diff
@@ -8,3 +8,4 @@ dekorate.kubernetes.ports[0].containerPort=8080
 dekorate.kubernetes.serviceType=LoadBalancer
 dekorate.kubernetes.configMapVolumes[0].volumeName=bar-volume
 dekorate.kubernetes.configMapVolumes[0].configMapName=foo-map
+dekorate.docker.group=dekorate.io
```
iocanel: Hi Jack! Thanks for taking the time to contribute!
Can you please provide some more details on how this makes image manipulation cleaner?

jbeaken: Hi, thanks for getting back! I upgrade from 0.7 to the 0.10.11 and the image property in the generated yaml file stopped using the **dekorate.kubernetes.group** property as a prefix. I think this is right, as it wasn't ideal behaviour as stated in https://github.com/dekorateio/dekorate/issues/409

I was using amazon ecr so wanted a prefix of [ACCOUNT_NUMBER].dkr.ecr.us-east-1.amazonaws.com

It wasn't clear how to achieve this, but eventually I discovered **dekorate.docker.group** (and version etc)

It would have saved me (and others?) a lot of time if this example (spring-boot-on-kubernetes-without-annotations) had made use of **dekorate.docker.group**

At the moment, the readme.md of this example states that the generated file includes  
`image: "hello-world/spring-boot-on-kubernetes-without-annotations:0.8-SNAPSHOT"`

but it doesn't, it actually (for me jbeaken) generates **jbeaken/spring-boot-on-kubernetes-without-annotations:0.10-SNAPSHOT** 

So I created this pull request to hopefully make it clearer. If you want to proceed, I can add tests etc.

thanks!

iocanel: We definitely, need to take action and make it clear for everyone.

I'd prefer if we had a dedicated example for that, maybe something like `spring-boot-with-custom-docker-image-name` or something like that and an entry in the documentation. Is this something you would like to add?

jbeaken: I should be able to do, as I don't know dekorate too well, though we will be using it in our next project. 

It would be helpful if i could use a 'base' example get me started, for example the `spring-boot-on-kubernetes-without-annotations `

Is that something you'd be happy with?

iocanel: That sound good, just scrap all other properties from application.properties and just keep `dekorate.docker.group`.

iocanel: Let's hold that thought for a while.

I am considering of completly wiping `group` our of `kubernetes`, `openshift` and `knative` configuration and just keep it around for image config like `docker` and `jib`. This will completlely demistify this.

In the place of `group` we can use `part-of` which is also the offcial recommendation of kubernetes.

jbeaken: sure thing, whatever you think best

iocanel: The last commit of #459 sums up what I have in mind. Feedback is more than welcome.

PR #472, opened 2020-03-24
fix: Decorator override-ability

iocanel:
This pull request deals with decorator ordering.
Also, it makes it possible to override image pull policy by explicilty providing a decorator.

### DIFF
--- a/core/src/main/java/io/dekorate/AbstractKubernetesHandler.java
+++ b/core/src/main/java/io/dekorate/AbstractKubernetesHandler.java
@@ -15,6 +15,7 @@
  */
 package io.dekorate;
 
+import io.dekorate.kubernetes.annotation.ImagePullPolicy;
 import io.dekorate.kubernetes.config.Annotation;
 import io.dekorate.kubernetes.config.AwsElasticBlockStoreVolume;
 import io.dekorate.kubernetes.config.AzureDiskVolume;
@@ -83,14 +84,6 @@ public abstract class AbstractKubernetesHandler<C extends BaseConfig> implements
    * @param config    The config.
    */
   protected void addDecorators(String group, C config) {
-    if (Strings.isNotNullOrEmpty(config.getServiceAccount())) {
-      resources.decorate(new ApplyServiceAccountNamedDecorator(config.getName(), config.getServiceAccount()));
-    }
-    resources.decorate(group, new ApplyImagePullPolicyDecorator(config.getImagePullPolicy()));
-
-    for (String imagePullSecret: config.getImagePullSecrets()) {
-      resources.decorate(new AddImagePullSecretDecorator(config.getName(), imagePullSecret));
-    }
 
     //Metadata handling
     resources.decorate(new AddVcsUrlAnnotationDecorator());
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/dekorate/AbstractKubernetesHandler.java
```diff
@@ -86,7 +87,10 @@ protected void addDecorators(String group, C config) {
     if (Strings.isNotNullOrEmpty(config.getServiceAccount())) {
       resources.decorate(new ApplyServiceAccountNamedDecorator(config.getName(), config.getServiceAccount()));
     }
-    resources.decorate(group, new ApplyImagePullPolicyDecorator(config.getImagePullPolicy()));
+
+    if (config.getImagePullPolicy() != ImagePullPolicy.IfNotPresent) {
```
geoand: Can't we just drop it completely?

I am probably wrong because I am thinking about this from a Quarkus perspective where the `ApplyImagePullPolicyDecorator` is always applied.

iocanel: No, we can't or we'll break it on dekorate side.

geoand: :+1:

geoand:
CI seems to be complaining

geoand:
```
[INFO] Running io.dekorate.examples.kubernetes.Feat458Test
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.591 s <<< FAILURE! - in io.dekorate.examples.kubernetes.Feat458Test
[ERROR] shouldHaveKubernetestLabelsKnativeYml  Time elapsed: 0.449 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: not <null>
```

core/src/main/java/io/dekorate/Resources.java
```diff
@@ -153,9 +153,16 @@ public void addCustom(String group, HasMetadata metadata) {
 
     Map<String, KubernetesList> resources = new HashMap<>();
 
-    if (groupDecorators.isEmpty()) {
-      groups.forEach((group, l) -> groupDecorators.put(group, globalDecorators));
-    }
+    System.out.println("Global decorators:" + globalDecorators.stream().map(d -> d.getClass().getSimpleName()).collect(Collectors.joining(" ")));
```
geoand: Do we need this?

core/src/main/java/io/dekorate/Resources.java
```diff
@@ -153,9 +153,16 @@ public void addCustom(String group, HasMetadata metadata) {
...
+    groups.forEach((group, l) -> {
+        if (!groupDecorators.containsKey(group) || groupDecorators.get(group).isEmpty()) {
+          groupDecorators.put(group, globalDecorators);
+        }
+    });
+
+    groupDecorators.forEach((group, decorators) -> {
+        System.out.println("Group: " + group + " decorator: " + decorators.stream().map(d -> d.getClass().getSimpleName()).collect(Collectors.joining(" ")));
```
geoand: Same here

core/src/main/java/io/dekorate/Resources.java
iocanel: No, we don't. I'll remove that.

PR #480, opened 2020-04-09
feat: Add a simple SessionReader impl for files

iocanel:
Need an non-apt implementation of a SessionReader.


geoand:
Where will this be used?

iocanel:
Wuarkus mostly.

PR #488, opened 2020-04-29
Make all loggers use the one created with the session

geoand:
We need to do this because otherwise other loggers won't
use the same logger


geoand:
@iocanel can we have this for the next release?

PR #492, opened 2020-05-07
Allow generating manifests outside target

iocanel:
This pull request makes it possible to write manifests outside of target by leveraging the property `dekorate.options.output-path`. 

The property was already available via annotations, but now it can be set from .yml/.property files too.

Also, this pull request moves geenrated metadata in they own special place `../.dekorate` relative to output classes directory.


geoand:
Question: Is this going to break any existing applications or integrations in any way?

iocanel:
> Question: Is this going to break any existing applications or integrations in any way?
I don't think so.

geoand:
Can you try it out for Quarkus for example?

PR #494, opened 2020-05-07
fix: Generated pvc now uses correct format

iocanel:


annotations/tekton-annotations/src/main/java/io/dekorate/tekton/annotation/TektonApplication.java
```diff
@@ -108,11 +108,11 @@
   String sourceWorkspaceClaim() default "";
 
   /*
-   * The size requirement of the generated PVC
+   * The size requirement of the generated PVC in gigs.
```
geoand, line 111: Maybe it should be megabytes instead?

iocanel: This needs to be configurable, but right now I need a quick fix.

PR #495, opened 2020-05-08
refactor: examples are now part of the release

iocanel:


iocanel:
> LGTM. The test is failing because of timeout with Openshift?

No matter what we do this happens from time to time!

aureamunoz:
Yes, yes, I know it, wanted just to be sure :-)

PR #518, opened 2020-06-01
fix (#517) optional is now excluded for env var source and key selector.

### LINKED ISSUE #517: Exclude 'optional' property from env sources and key selectors
Related to https://github.com/quarkusio/quarkus/issues/9561.

I think that we should generally exclude all fields baring default values. 
This way, we will cover cases where compatibility is comprised, without restricting the user (as we will only reference properties that the user configuration directly affects).

iocanel:
Fixes #517 

As it was not possible to address the issue on Serialization level, we address just the case with `optional` on decorator level.


geoand:
What's up with CI?

iocanel:
> What's up with CI?

A timeout during the s2i build.

Issue #519, opened 2020-06-02
Decouple logic from annotations


iocanel:
The dekorate logic is coupled with the annotations. This prevents us from being able to separate the actual annotations, processors and logic.

Decoupling is a prerequisite of #516, as we will need to split existing modules into logic and processors. Ideally, after the split the logic modules should not include the annotations, as it could allow users to shoot their feet.

PR #520, opened 2020-06-02
Decouple annotations from logic

iocanel:
Fixes: #519

The main changes of the pull request are the following:

- Generators now return config class instead of annotation class.
- Code that references annotations moved from `Generator` class to `AnnotationProcessor` classes


geoand:
I won't have time to review before Thursday, so you might want to ping someone else if you need it before then

aureamunoz:
I'll take a look!

iocanel:
> LGTM!
> It's maybe a stupid question but, why did you do this refactor @iocanel ?

I want the flexibility to be able to split the annotation processors from the rest of the code. Right now, they are tightly coupled.
**Bonus point**: We also align the execution path between annotation/annotationless modes.

Issue #536, opened 2020-06-16
Implement Info readers for nodejs projects


iocanel:
While, dekorate itself is geared towards java projects, the dekorate-cli could be used against any project. It would be nice if internally we could support `nodejs` projects.

PR #541, opened 2020-06-16
feat (#536) Introduce nodejs framework module.

iocanel:
Fixes: #536


frameworks/nodejs/src/main/java/io/dekorate/nodejs/NodeInfoReader.java
```diff
@@ -0,0 +1,175 @@
...
+      .withBuildTool(NPM)
+      .withBuildToolVersion(getVersion(root))
+      .withOutputFile(root.resolve("app.js"))
+      .withClassOutputDir(null)
+      .withResourceDir(null)
+      .build();
+  }
+
+  /**
+   * Parse build.gradle and read the jar configuration as a {@link Map}.
```
geoand, line 110:
```suggestion
   * Parse package.json and read the jar configuration as a {@link Map}.
```
Also it would be nice to have a comment on what exactly is parses

iocanel:
@geoand: Applied feedback. Are we good?

Issue #545, opened 2020-06-17
Port path is ignored


iocanel:
Given:

```
dekorate.kubernetes.ports[0].name=http
dekorate.kubernetes.ports[0].path=/hello
dekorate.kubernetes.expose=true
```
I expect an ingress generating that uses the path `/hello`.
What acutally happens is that the path used is `/`.

From a quick debug I see that dekorate internally uses 3 instances of port:
- one completely empty
- one with right port and path
- one with right port and no path

I would expect two (defaults & provided) that should be merged into one.

We need to check why this doesn't happen.

PR #546, opened 2020-06-17
Fix (#545) Ports are now properly merged.

iocanel:
Fixes: #545 

This pull request fixes some issues with the a bean merger, that is used to merge configuration together (property, annotation and framework configurration).

The main issue, is that when merging arrays, lists and set we didn't take ihertitence into account. This PR fixes that and also improves on the way we counfugre ports.

### DIFF
--- a/core/src/main/java/io/dekorate/kubernetes/configurator/ApplyContainerPort.java
+++ b/core/src/main/java/io/dekorate/kubernetes/configurator/ApplyContainerPort.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2018 The original authors.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * 
+**/
+
+package io.dekorate.kubernetes.configurator;
+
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/dekorate/kubernetes/configurator/ApplyContainerPort.java
```diff
@@ -0,0 +1,54 @@
...
+    this(containerPort, Arrays.asList(names));
+  }
+
+  @Override
+  public void visit(BaseConfigFluent<?> config) {
+    Predicate<PortBuilder> predicate = p -> names.contains(p.getName());
+    if (config.hasMatchingPort(predicate)) {
+      config.editMatchingPort(predicate).withContainerPort(containerPort).endPort();
+    } else {
+      String name = names.get(0);
```
geoand, line 47: Are we sure this always exists?

core/src/main/java/io/dekorate/utils/Beans.java
```diff
@@ -167,8 +166,9 @@
     try {
       Class originClass = origin.getClass().getComponentType();
       List result = new ArrayList<>();
-      List originList = Arrays.asList((Object[])origin);
-      List overrideList = Arrays.asList((Object[])override);
+      //We wrap the Arrays.asList() in an ArrayList so that we can mutate it.
```
geoand, line 169: Inception :P

geoand:
Would be nice to see a test that shows that #545 is indeed addressed. Is that possible?

core/src/main/java/io/dekorate/kubernetes/configurator/ApplyContainerPort.java
iocanel: As this is only used in one place, yes we are sure.
It's a bad idea nonetheless. 
Fix incoming.

iocanel:
> Would be nice to see a test that shows that #545 is indeed addressed. Is that possible?

Yeap!

iocanel:
Added test!

Issue #559, opened 2020-06-26
Allow using docker and jib on openshift


iocanel:
Currently, if you use openshift, you have to go the s2i route.
It should be possible to use docker or jib if needed.

Related to: https://github.com/quarkusio/quarkus/issues/9663

Issue #560, opened 2020-06-26
Jib build service doesn't work correctly on when used with testing modules.


iocanel:
The JibBuildService requires an instance of JibConfig.
That prevents it from working properly when used with the junit5 extension, that currently feed a generic instance of image configuration.

The service should optionally use a JibBuildConfigm.

PR #561, opened 2020-06-26
Allow openshift with non s2i builds

iocanel:
Resolves: #559 #560


geoand:
I'll check it over the weekend or Monday

PR #574, opened 2020-07-03
feat (#572) Add support for cluster local services.

iocanel:
Resolves: #572


iocanel:
> So this changes the default behavior, right?

No, it doesn't. We also change the default values so that services are still exposed ootb.

geoand:
Okay, thanks

Issue #593, opened 2020-07-20
branching strategy


Ladicek:
I'm trying to make a little change in Dekorate, and found that I totally can't comprehend all the branches here.

On `master`, the README says:

```
- master (active development)
- 1.0.x  (important bug fixes)
- 0.12.x (new features + bug fixes)
```

But GitHub shows 5 branches: `0.11.x`, `0.12.x`, `1.0.x`, `2.0.x`, `master`. The change I'd like to do is in Quarkus context, and Quarkus has Dekorate 0.12.6, so I assume I want to do a PR against `0.12.x`. Or perhaps both `0.12.x` and `master`? Or perhaps all of `0.12.x`, `2.0.x` and `master`?

For me, who doesn't work regularly on Dekorate, the current branches structure is immensely confusing. I'd appreciate if this was streamlined. (How come `1.0.x` is older and slower than `0.12.x`? Why is there a separate `2.0.x` branch, when `master` includes some/all of the refactorings as well?)

iocanel:
Thanks for bringing this up @Ladicek. That's great feedback!

`master` is the branch that we use for development. Everything goes there and we cherry-pick to wherever we need to.

We use two different release branches:

`1.0.x` is the slow paced branch (only bug fixes since 1.0.0).
`0.12.x` is the fast paced branch (new features / bug fixes, etc).

I agree that its totally confusing that the 0.12.x branches moves faster than the 1.0.x one. So, the idea is to abandon 0.12.x and use 2.0.x as the fast paced branch, which will hopefully make things feel more natural.

Ladicek:
How about just keeping `1.0.x` branch for the slow branch, and cut 2.x releases from `master`? Are there really changes going to `master` that are not suitable for inclusion into 2.x?

iocanel:
That's the plan really. But we are not ready to align master with 2.0.x yet. And I want to be able to release 0.12.x at will till then.

Ladicek:
Cool, thanks!

iocanel:
Let me update the doc, so that its as less confusing as possible.

PR #594, opened 2020-07-23
role binding names

Ladicek:


Ladicek:
I'm creating the PR against the 0.12.x branch, because #593.

iocanel:
For future reference. Let's keep PRs pointing to master and I will backport when needed

iocanel:
... unless of course the fix doesn't make sense in Master.

PR #601, opened 2020-07-29
chore: update orb to 0.4.0

aureamunoz:


aureamunoz:
WDYT @iocanel @metacosm ?

metacosm:
I don't really have an opinion on this. Do we really need that update? What does it bring that the current version doesn't?

aureamunoz:
I don't know exactly but there is only one tag 0.4.0 so thought it would be good idea to use this one? 
https://github.com/snowdrop/circleci-orbs/tags

iocanel (approved):
LGTM

PR #641, opened 2020-09-25
Fix Issue #640 Gradle resource dir not set properly

### LINKED ISSUE #640: Gradle resourceDir not set properly
When using gradle project type resourceDir is set to classOutputDir thus application properties are not applied

dswiecki:
Resolves #640

### DIFF
--- a/core/src/main/java/io/dekorate/project/BuildInfo.java
+++ b/core/src/main/java/io/dekorate/project/BuildInfo.java
@@ -47,22 +47,6 @@ public class BuildInfo {
    * @param buildToolVersion The build tool version
    * @param outputFile The output file (the path to the actual jar, war etc).
    * @param classOutputDir The resource output directory (e.g. target/classes, build/classes/main/java etc).
-   */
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/dekorate/project/BuildInfo.java
```diff
@@ -37,22 +37,6 @@ public BuildInfo() {
     this(null, null, DEFAULT_PACKAGING, DEFAULT_BUILD_TOOL, null, null, null, null);
   }
 
-  /**
```
iocanel, line 40: This might be used by downstream projects, so I'd like to bring this back.

iocanel (changes requested):
Overall, it looks good. 

The only thing I'd like to change is to bring back the constructor that got deleted.

iocanel (approved):
I went on to add the method back, then I felt it might be better this way as in the long run it might cause more issues than it will solve.

So, I'll accept it as is.

Issue #553, opened 2020-06-22
Support for resource level annotations


kameshsampath:
It will be good to have resource level annotations configurable via properties. E.g. let us say we have a `dekorate.kubernetes.expose=true`, then doing something like 

```
dekorate.kubernetes.expose.annnotations[0].key=foo
dekorate.kubernetes.expose.annnotations[0].value=bar
```

Should add the annotations to expose related resources such as *ingress*

iocanel:
I would prefer if we used `ingress` instead of `expose` unless these annotations are also applicable to Route too.

Also, do you think that we should expose the annotation itself or a property describes the intent, e.g. `dekorate.kubernetes.ingress.rewrite-target`?

kameshsampath:
@iocanel , I initially thought about using specific k8s resource kind, then wondering if we are complicating :( .. so thought to make it general as annotations.

kameshsampath:
>Also, do you think that we should expose the annotation itself or a property describes the intent, e.g. dekorate.kubernetes.ingress.rewrite-target?

That sounds like a good idea too

iocanel:
I have been pondering for a solution around this and TBH, I haven't been able to find an elegant technical soluton.

The best thing that pops to my mind, is to modify @Label and @Annotation in order to include a field say `resourceKinds`.

This way one could 

```java
@KubernetesApplication(name="myapp", labels=@Label(name="foo", value="bar", rerourceKinds=["Service", "Ingress"])
```

or in properties;

```
dekorate.kubernetes.labels[0].key=foo
dekorate.kubernetes.labels[0].value=bar
dekorate.kubernetes.labels[0].resource-kinds=Service,Ingress
```

I  think that this is pretty close to what @kameshsampath needs and it should also be compatible with what have in downstream projects. @geoand wdyt?

geoand:
I like the idea!

PR #644, opened 2020-10-01
Feat 553 resource level annotations

iocanel:
Resolves: #553 

This pull request introduces a new fields `kinds` on `@Label` and `@Annotation`. The use of these fields will limit the application of Labels/Annotations only to the listed kinds. If omitted, all kinds are asumed.

### DIFF
--- a/annotations/halkyon-annotations/src/main/java/io/dekorate/halkyon/adapter/ComponentConfigAdapter.java
+++ b/annotations/halkyon-annotations/src/main/java/io/dekorate/halkyon/adapter/ComponentConfigAdapter.java
@@ -53,8 +53,8 @@ public class ComponentConfigAdapter {
             i.secret(),
             i.configmap(),
             i.field())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Env[0]),
-        Arrays.asList(instance.labels()).stream().map(i -> new io.dekorate.kubernetes.config.Label(i.key(),
-            i.value())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Label[0]),
+                                                                                     Arrays.asList(instance.labels()).stream().map(i -> new io.dekorate.kubernetes.config.Label(i.key(), i.value(),
+                                                                                                                                                                                i.kinds())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Label[0]),
         instance.buildType(),
         instance.remote(),
         Arrays.asList(instance.provides()).stream().map(i -> new io.dekorate.halkyon.config.CapabilityConfig(null,
@@ -121,7 +121,8 @@ public class ComponentConfigAdapter {
   private static Label getLabel(Map i) {
     return new Label(
         (String) i.getOrDefault("key", null),
-        (String) i.getOrDefault("value", null));
+        (String) i.getOrDefault("value", null),
+        (String[]) i.getOrDefault("kinds", new String[0]));
   }
 
   private static Env getEnv(Map i) {
--- a/annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
+++ b/annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
@@ -25,9 +25,7 @@ public class KnativeConfigTest {
         .withName("myapp")
         .withVersion("1.0.0")
         .withExpose(true)
-        .addNewLabel("key1", "val1")
+        .addNewLabel("key1", "val1", new String[0])
         .build();
-
   }
-
 }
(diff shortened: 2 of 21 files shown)


annotations/halkyon-annotations/src/main/java/io/dekorate/halkyon/adapter/ComponentConfigAdapter.java
```diff
@@ -53,8 +53,8 @@ public static ComponentConfigBuilder newBuilder(HalkyonComponent instance) {
             i.secret(),
             i.configmap(),
             i.field())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Env[0]),
-        Arrays.asList(instance.labels()).stream().map(i -> new io.dekorate.kubernetes.config.Label(i.key(),
-            i.value())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Label[0]),
+                                                                                     Arrays.asList(instance.labels()).stream().map(i -> new io.dekorate.kubernetes.config.Label(i.key(), i.value(),
+                                                                                                                                                                                i.kinds())).collect(Collectors.toList()).toArray(new io.dekorate.kubernetes.config.Label[0]),
```
geoand, line 57: Formatting looks messed up

annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
```diff
@@ -25,9 +25,7 @@ public void testGeneratedCode() {
         .withName("myapp")
         .withVersion("1.0.0")
         .withExpose(true)
-        .addNewLabel("key1", "val1")
+        .addNewLabel("key1", "val1", new String[0])
```
geoand, line 28: Do we really need to do this? Can't the existing `addNewLabel` continue to exist?

geoand (commented):
I would like to know what this means in terms of backward compatibility

annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
iocanel: No, it can't.

Possible alternatives:

```java
.addNewLabel().withKey("key1").withValue("val1").endLabel()
```

Given that this change is not user facing, I'd say its no biggie.

annotations/halkyon-annotations/src/main/java/io/dekorate/halkyon/adapter/ComponentConfigAdapter.java
iocanel: Weird as we now use a maven plugin for formatiing. Will look closer.

iocanel:
> I would like to know what this means in terms of backward compatibility

There will be no backwards compatibility, but will be forward compatibility (partially).
By partially I mean that dekorate user can upgrade without issues.
Downstream projects may need to perform minor changes.

So, most probably this will require a 0.13.0 release.

annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
geoand: Sure yeah, it just looks ugly

geoand:
CI seems to be failing

annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
iocanel: autoformatting doesn't work as expected.

iocanel: @geoand: The main thing with these methods is that they are generated. 

The logic is the following:
`if the object is flat and has less than five 5 fields generate a with method with all fields`.

In theory we could create a method for each possible field combination. In the current case it would look like:

```
withNewLabel(String key);
withNewLabel(String key, String value); 
withNewlabel(String key, String value, String[] kinds);
```
However, this would mean that we would also get methods that don't make sense. 
Now, since to original object is an annotation and fields without default values are considered mandatory we could possibly work soemthing out.

I think that this would be one or two days work. and I'd like to avoid it at the moment.
Created: https://github.com/sundrio/sundrio/issues/177

annotations/halkyon-annotations/src/main/java/io/dekorate/halkyon/adapter/ComponentConfigAdapter.java
iocanel: Formatter is not used by default. I will address this in a separate issue.

annotations/knative-annotations/src/test/java/io/dekorate/knative/config/KnativeConfigTest.java
iocanel: @geoand if this sounds reasonable can we approve this?

geoand (approved):

Issue #620, opened 2020-09-15
Provide and document an example where the full image name is provided by the user.


iocanel:

cmoulliard:
I suppose that `Provide` means that you will add and support another new parameter `imageName` for the Kubernetes or OpenshiftApplication annotation able to override the image name with the name provided

Example

```
@KubernetesApplication(
        name = "hello-world-fwless-k8s",
        ports = @Port(name = "web", containerPort = 8080),
        serviceType = ServiceType.NodePort,
        expose = true,
        imagePullPolicy = ImagePullPolicy.Always,
        imageName = localhost:5000/hello-world-fwless:1.0-SNAPSHOT)
```

PR #645, opened 2020-10-01
fix (#620): image property is not properly read. Add doc.

iocanel:
Resolves: #620


aureamunoz (approved):
Cool! I will use it with the fwless app

PR #653, opened 2020-10-08
Kubernetes Container Resources

Vinche59:
As stated here : https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

The generated resources for memory limits/requests should be "memory" and not "mem"

Signed-off-by: Vincent Sourtin <sourin-v@bridgestone-bae.com>


iocanel (approved):
Great catch!

PR #666, opened 2020-10-21
refactor: Labels and annotations are now only added to named

iocanel:
resources (vs all resources).

As we are incosistnent on where we apply label and annotation decorator, this pull request limits that to just named resources. In other words resources directly generated by us.


iocanel:
@geoand: I know that you are snowed. It's more like an FYI rahter than an actual request.

geoand:
Judging by the PR number, it's pretty evil 😂

iocanel:
> Judging by the PR number, it's pretty evil

Right! It's vi vi vi! Super evil!

PR #209, opened 2019-05-23
[WIP] feat: Add support for generating custom resource model classes.

iocanel:

PR #674, opened 2020-11-10
Update #209 to bring it in line with latest version

metacosm:
See https://github.com/dekorateio/dekorate/pull/209.
This also updated the kubernetes-client to 4.13.0.


metacosm:
Obsolete.

PR #685, opened 2020-12-10
feature: use default registry if registry value absent

aureamunoz:
I think we should use `DEFAULT_REGISTRY` if no value provided for registry, as done in OpenShiftHandler. WDYT, @iocanel ?


iocanel:
> I think we should use `DEFAULT_REGISTRY` if no value provided for registry, as done in OpenShiftHandler. WDYT, @iocanel ?

In most cases, that should be fine, but it would break environments like minikube, where no push to registry is actually required.

iocanel (changes requested):
I think that before we get this in, we need a module that would differenciate the defaults for minikube.

iocanel:
Given that we mered #792 I think we can close thins one.

PR #701, opened 2021-01-11
DekorateException , Infinite loop, and package name issues reproduction

wtrocki:
Generally making changes inside sample app ends up with issues:
io.dekorate.DekorateException or StackOverflow.

This PR will contain number of cases to reproduce errors that will be linked to the underlying issue.


iocanel:
Issue has been fixed. Closing.

Issue #700, opened 2021-01-08
Fatal error compiling: java.lang.StackOverflowError thrown by generation process


wtrocki:
Running Zookeper example

```
[INFO] Generating manifests.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.947 s
[INFO] Finished at: 2021-01-08T14:02:28Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project Dekorate :: Examples :: POJO to Custom Resource :: Zookeeper example (imaginary): Fatal error compiling: java.lang.StackOverflowError -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
```

iocanel:
@wtrocki: Can you re-run the example with `-e` and provide a full stack trace?

wtrocki:
Stack trace is not useful. It ends up with infinite loop.

```
  at io.dekorate.crd.visitor.AnnotatedPropertyPathDetector.visit (AnnotatedPropertyPathDetector.java:71)
    at io.dekorate.crd.visitor.AnnotatedPropertyPathDetector.visit (AnnotatedPropertyPathDetector.java:33)
    at io.sundr.builder.BaseFluent.acceptInternal (BaseFluent.java:149)
    at io.sundr.builder.BaseFluent.accept (BaseFluent.java:139)
    at io.dekorate.crd.visitor.AnnotatedPropertyPathDetector.lambda$visit$2 (AnnotatedPropertyPathDetector.java:77)
    at io.dekorate.crd.visitor.AnnotatedPropertyPathDetector$$Lambda$283/0000000000000000.accept (Unknown Source)
    at java.util.stream.ForEachOps$ForEachOp$OfRef.accept (ForEachOps.java:183)
    at java.util.stream.ReferencePipeline$2$1.accept (ReferencePipeline.java:177)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining (Arr^Cm io.sundr.builder.BaseFluent.accept (BaseFluent.java:139)
[...]

Issue #398, opened 2019-11-04
Bulding dekorate using OpenJ9 jdk fails


iocanel:
The issue is that generated code includes ibm specific classes that compiler doesn't really like.

These classes should never end up in the generated code. Only explantation for them being there is that the OpenJ9 compiler sneaks references to them in certain types, causing `sundrio` to belive that they are referenced by the code and thus generating import entries.

PR #702, opened 2021-01-12
Upgrade kubernetes-client to 5.0.0 and sundrio to 0.24.0

iocanel:
This pull request performs the following changes:
- kubernetes-client -> 5.0.0
- sundrio -> 0.24.0
- fix compilation issues with openj9
- fix stack overflow issues on certain conditions when using openj9

--
Most of the code changes are related to:
- Replacement of sundrio Predicate by `java.util.function.Predicate`
- Removal of Doneables (no longer generated by the client).
- Package chnage affecting `Ingress` and `CustomResourceDefinition`

Resolves: #398 
Resolves: #700


geoand (approved):
+1 assuming tests pass :)

annotations/crd-annotations/src/main/java/io/dekorate/crd/decorator/AddAdditionPrinterColumnDecorator.java
```diff
@@ -18,9 +18,9 @@
 package io.dekorate.crd.decorator;
 
 import io.dekorate.utils.Strings;
-import io.fabric8.kubernetes.api.builder.Predicate;
-import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceColumnDefinitionBuilder;
-import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinitionVersionFluent;
+import java.util.function.Predicate;
+import io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceColumnDefinitionBuilder;
```
metacosm, line 22: Should probably use `v1` version instead of `v1beta1`…

iocanel: At some point we should probably do that, but I think that for now we should keep the version we already used.
Going to `v1` is not a trivial task and will possibly cause lot of issues.

PR #703, opened 2021-01-12
chore: Add link do contributing documentation

wtrocki:
## Motivation

Contributing.md is standard for github repositories. I did not want to move file from original assets folder as there is value of keeping the docs in the same folder. That is why we adding contributing guide but linking to the other document.


iocanel (approved):
LGTM

PR #704, opened 2021-01-12
chore: Add note for usage

wtrocki:
Adding extra note to the readme to consider using 2.0.0-alpha.
I did not wanted to use exact version as this might quickly oudate


iocanel (approved):
LGTM

PR #717, opened 2021-01-25
Make CRD module standalone

### LINKED ISSUE #716: Make crd-annotations a standalone module
The goal here is to make `crd-annotations` a module that can be used with or without the rest of the dekorate stauff.

In order to do so, we will have to replicate some concepts inside the module and discard some other.

### Decorator 
We need the Decorator concept along with some core dekorators like `NamedResourceDecorator` etc.
It's important that we also keep the concept of sorting decorators as we do make use of this capability.

### Serialization
Currently dekorate has its own `Serialization` uitl that is slightly different than the kubernetes-client one.
It's tailor made to resource generatin and we do need to keep it around.

## Configurator
We only use a couple and they could be easily replaced with regular Visitors.

## Generator / Hanlder

In regular dekorate modules, we don't process the configuration when we actually get it (from Annotation  Processors).
The `Generator` stores it into the `Session` and when the `Session` closes, we do build the configuration, by applying configurators etc and then pass it to the appropriate handler. This is done to support multiple modules  (e.g. kubernetes, openshift, docker etc).

For this use case, we can completely ignore those and have to AnnotationProcessor directly call the Handler to process the configuration and then the `Serialization` to generate the actual code.

metacosm:
Fixes #716

### DIFF
--- a/annotations/crd-annotations/pom.xml
+++ b/annotations/crd-annotations/pom.xml
@@ -70,6 +70,18 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-model-common</artifactId>
+      <version>5.0.0</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-model</artifactId>
+      <version>5.0.0</version>
+    </dependency>
+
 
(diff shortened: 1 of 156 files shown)


iocanel (changes requested):
It might not be obvious by the issue description, but we don't need to carry over the Session. We can go difrrectly from `apt`  -> `handler` and just use provided decorators.

PR #724, opened 2021-02-03
Fixed integer schema typo

ppatierno:
This trivial PR fixes a typo in the JSON integer schema that is the cause of a malformed CRD using "ineger" instead of "integer" type.


aureamunoz (approved):
LGTM. Thank you!!

metacosm (approved):

Issue #671, opened 2020-11-06
Maven stuck during compilation when launched in debug mode


aureamunoz:
It seems like maven goes into some lock while compiling classes, it remains in this state.
To reproduce run the following command:
```
mvnDebug clean install -Pwith-examples -pl :kubernetes-example-with-custom-image-name -Dinvoker.test=KubernetesExampleTest`
```

When the test is performed without debug enabled everything is ok.

I have added the following configuration to the maven compiler plugin in order to see more traces, but I can not see the actual cause of the blocking.

```
<configuration>
            <compilerArgs>
              <arg>-verbose</arg>
              <arg>-Xlint:all,-options,-path</arg>
            </compilerArgs>
        </configuration>
```

aureamunoz:
it seems that all mvnDebug really does is setting the environment variables needed to start java in debug mode.
so the new process inherits the env from the parent process and even if we don't use mvnDebug the environment variables are there, causing the new java process to also start in debug mode. So, we need to strip the new process of the environment variables

PR #729, opened 2021-02-18
fix: unlock mvnDebug process by removing the debug env vars

aureamunoz:
Finally!
Should we add some test? I'll let you take a look @iocanel, let me know if something is needed


iocanel (approved):
LGTM. Not sure how we could test it.
When CI gives us a green light we can merge.

aureamunoz:
Fixes #671

PR #730, opened 2021-02-19
Introduce vcs options

iocanel:
The idea is to provide the following features:

- ability to select remote that will fuel vcs annotations on resources
- introduce feature flag for vcs url sanitization


iocanel:
@metacosm: This pull request changes the default behavior of dekorate (vcs url are no longer sanitized by default), which is something you added.

@geoand @aureamunoz: some of the modified files have been recently edited by you.

geoand (approved):
LGTM

Issue #727, opened 2021-02-17
Align or warn when parameters in app/image config differ.


iocanel:
Currently, the user is free to set parameters like:
- group
- name 
- version

in the application configuration e.g. vi @KubernetesApplication or `dekorate.kubernetes.xxx`.
However, user needs to align those with the image configuration.

Sometimes, this is not obvious to users. The problems caused by this are really hard to reproduce and can cause great frustration.

We need to either warn users or automatically align configurations ourselves when this happens.

I tend to slightly favor the later. We should check if explicit values have been specified for the image configuration. 
If image configuration has no explicit values, then we should pickup whatever is specified in the application config.
If both have been specified and are conflicting we need to throw an error.

PR #732, opened 2021-02-22
Align group, name and version

iocanel:
Resolves: #727

The pull request does the following:

- Separates addition of configurers from configuration (allows keeping track of what explciitly/implictly configured).
- Apply single application configuration to image configuration with missing explicit values for group, name & version.
- Throw error when there are conficting explict values

### DIFF
--- a/annotations/docker-annotations/src/main/java/io/dekorate/docker/config/DefaultDockerBuildConfigGenerator.java
+++ b/annotations/docker-annotations/src/main/java/io/dekorate/docker/config/DefaultDockerBuildConfigGenerator.java
@@ -18,19 +18,20 @@
 package io.dekorate.docker.config;
 
 import io.dekorate.ConfigurationRegistry;
+import io.dekorate.WithProject;
 import io.dekorate.config.DefaultConfiguration;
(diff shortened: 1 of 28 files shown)


aureamunoz (changes requested):
I have done some minor comments. I'm not sure about this behaviour: ` If both have been specified and are conflicting we need to throw an error.` but I don't have an opinion strongly against it either, I just think that if an user set image configuration explicitly, why not pick it up? Why should we compare it with the application config? Does it make sense? I'm wondering about it. 

And yes, as already mentioned in chats, it would need tests.
Docs are very clair IMO.

iocanel:
> I have done some minor comments. I'm not sure about this behaviour: ` If both have been specified and are conflicting we need to throw an error.` but I don't have an opinion strongly against it either, I just think that if an user set image configuration explicitly, why not pick it up? Why should we compare it with the application config? Does it make sense? I'm wondering about it.

I think that you are right. Requiring absolute alignment is maybe too strict. I changed the behavior so that we just log a debug message, if this case is encountered.

aureamunoz (approved):

PR #734, opened 2021-02-24
refactor: ServiceBinding is no longer generated

iocanel:
Now that @Crd has been removed we need to make sure that it's no longer used in servicebingind-annotations


yharish991 (commented):
LGTM

yharish991:
oops looks like its already merged

Issue #706, opened 2021-01-14
Property configuration always overwrites values for primitive properties


iocanel:
The problem seems to be that the generated config objects, are generated with default values.
That works when they represent annotation configuration, but they do 'taint' the property configuration.

This leads to isses when it comes to merging of configs that include primitives.
For example setting the number of replicas via annotations, will be ignored if any property configuration is found, even if its irrelevant to replicas.

aureamunoz:
After spend a little time on this, yes the problem is that [we are always returning the value from the `higher priority configuration`](https://github.com/dekorateio/dekorate/blob/a958a8bb7c032b0e6ac15e5cf08426943c78a4c6/core/src/main/java/io/dekorate/utils/Beans.java#L77) in case of primitives or enums. I don't find the way to detect which is the conf that has been setted by user so what should we do? document clairly this case? Do you have an idea @iocanel ?

PR #737, opened 2021-03-04
fix (#706): generated config now uses autoboxing and mutable types

iocanel:
Resolves: #706

### DIFF
--- a/core/src/main/java/io/dekorate/utils/Generators.java
+++ b/core/src/main/java/io/dekorate/utils/Generators.java
@@ -16,9 +16,12 @@
 package io.dekorate.utils;
 
 import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 public class Generators {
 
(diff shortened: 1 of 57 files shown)


annotations/openshift-annotations/src/main/java/io/dekorate/openshift/config/OpenshiftConfigGenerator.java
```diff
@@ -39,6 +39,7 @@ default String getKey() {
   }
 
   default void addAnnotationConfiguration(Map map) {
+    System.out.println(map);
```
aureamunoz, line 42: Same remark, can we remove this?

core/src/main/java/io/dekorate/utils/Generators.java
```diff
@@ -39,6 +42,84 @@ private static Field findField(Class c, Class origin, String field)
...
+            }
+          } else {
+            if (fieldType.getComponentType().equals(Boolean.class)) {
+              System.out.println("boolean array mapping");
+              map.put(key, Arrays.stream(String.valueOf(value).split("\\s*,\\s*"))
+                      .map(Boolean::parseBoolean)
+                      .collect(Collectors.toList())
+                      .toArray(new Boolean[]{}));
+            } else if (fieldType.getComponentType().equals(Short.class)) {
+              System.out.println("short array mapping");
```
aureamunoz, line 77: Should we, maybe later in a new PR, log with `LOGGER`?

aureamunoz (changes requested):
This is great!!! I have taken a look on it and make just a few minor remarks. I have tested with my branch where I'm overriding `replicas` and it works!

PR #750, opened 2021-04-14
fix: parse Maven version in an ANSI escape sequence safe way

### LINKED ISSUE #749: "Unknown maven version output format" when output uses ANSI escape sequences
Over at https://github.com/quarkusio/quarkus/pull/16249 I'm seeing:
```
[WARN] Unknown maven version output format. Expected at least one line. Falling back to: 3.6.3!
```
while running `mvn clean verify -f integration-tests/kubernetes/quarkus-standard-way/`

_Edit: Not related to Maven 3.8.1, see https://github.com/dekorateio/dekorate/issues/749#issuecomment-819831585_

famod:
Fixes #749

### DIFF
--- a/core/src/main/java/io/dekorate/utils/Maven.java
+++ b/core/src/main/java/io/dekorate/utils/Maven.java
@@ -19,8 +19,8 @@ package io.dekorate.utils;
 
 import java.io.ByteArrayOutputStream;
 import java.nio.file.Path;
-import java.util.Arrays;
-import java.util.Optional;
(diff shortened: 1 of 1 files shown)


iocanel (approved):
LGTM

iocanel:
Thanks a lot @famod !!!

famod:
Yw!

FTR, in case the Maven guys decide that e.g. "Apache" gets another color than "Maven" in that output then this will fail again. But I don't think that'll happen. 😉

PR #751, opened 2021-04-20
improvement: use new jackson yaml formating version

dswiecki:
This uses new jackson-dataformat-yaml version which intends arrays elements


iocanel (approved):
LGTM

PR #753, opened 2021-04-27
Invalid logger breaks current application flow

pjgg:
Instead of throw an `IllegalStateException` exception we could log an `error` msg and swap to std out.

Related issue: https://github.com/quarkusio/quarkus/issues/15953

### DIFF
--- a/core/src/main/java/io/dekorate/logger/AnsiLogger.java
+++ b/core/src/main/java/io/dekorate/logger/AnsiLogger.java
@@ -20,6 +20,9 @@ package io.dekorate.logger;
 import static org.fusesource.jansi.Ansi.*;
 import static org.fusesource.jansi.Ansi.Color.*;
 
+
+import java.io.FileDescriptor;
+import java.io.FileOutputStream;
 import java.io.PrintStream;
 
 import org.fusesource.jansi.*;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/dekorate/logger/AnsiLogger.java
```diff
@@ -79,7 +82,22 @@ public void error(String message) {
 
   private void check() {
     if (stream == null) {
-      throw new IllegalStateException("AnsiLogger requires a PrintStream instance.");
+      stream = new PrintStream(new FileOutputStream(FileDescriptor.out));
```
iocanel, line 85: I wouldn't use Asciiart for something that is actually an error message.

iocanel (changes requested):
Thanks for the pull request.
I have a minor aesthetic change request. Other than that it looks good.

iocanel (approved):
LGTM

geoand:
@iocanel can we get a release of this that we then incorporate in Quarkus?

iocanel:
Will try to release later tonight or tomorrow morning!

geoand:
Thanks!

iocanel:
@pjgg: I have some `just a little` late questions:

- Does this pull request solve the issue with quarkus (Have you verified the fix)?
- Do you think that we should also bump jansi to a later version ?
- Why do you use `new FileOutputStream(FileDescriptor.out)` vs `System.out`?

PR #765, opened 2021-06-25
fix: Ingress now use `/` instead of blank path

iocanel:


cmoulliard:
This changes will break the compatibility with kubernetes <= 1.19. What do you plan to do for users using ingress with older releases ? @iocanel

iocanel:
> This changes will break the compatibility with kubernetes <= 1.19. What do you plan to do for users using ingress with older releases ? @iocanel

Compatibility was broken, in dekorate 2.0.0 where we moved to Ingress v1.
People, that are using <= 1.19 should stay with dekorate 1.x.

cmoulliard:
> Compatibility was broken, in dekorate 2.0.0 where we moved to Ingress v1.
> People, that are using <= 1.19 should stay with dekorate 1.x.

Is it documented and where ? It is very important that the developers know such a compatibility matrix and versions to be used ;-)

PR #777, opened 2021-08-11
Issue 775. Port numbers generated are not unique for the Kubernetes service

### LINKED ISSUE #775: The kubernetes service manifest generated contains duplicate values for port HTTP and HTTPS
## Issue

The kubernetes service manifest generated contains duplicate values for port HTTP and HTTPS using either dekorate 2.3.2 or 2.3.3 

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    app.openshift.io/vcs-url: <<unknown>>
  labels:
    app.kubernetes.io/name: springboot-dekorate-annotation
    runtimedb: mysql
    app.kubernetes.io/version: 2.0.0-SNAPSHOT
    app.openshift.io/runtime: spring-boot
  name: springboot-dekorate-annotation
spec:
  ports:
  - name: http
    port: 80
    targetPort: 8080
  - name: https
    port: 80
    targetPort: 8443
```

The following error is reported during the S2I deployment

```
The Service "springboot-dekorate-annotation" is invalid: spec.ports[1]: Duplicate value: core.ServicePort{Name:"", Protocol:"TCP", Port:80, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:""}, NodePort:0}
```

when the class is annotated as such

```
import io.dekorate.openshift.annotation.OpenshiftApplication;
import io.dekorate.s2i.annotation.S2iBuild;
import io.dekorate.kubernetes.annotation.Probe;
import io.dekorate.kubernetes.annotation.Label;
import io.dekorate.kubernetes.annotation.Port;
import io.dekorate.kubernetes.annotation.Env;

@OpenshiftApplication(name="springboot-dekorate-annotation",
                      expose=true,
                      replicas=2,
                      livenessProbe=@Probe(httpActionPath="/actuator/health",initialDelaySeconds=60,periodSeconds=3,timeoutSeconds=1),
                      readinessProbe=@Probe(httpActionPath="/actuator/health",initialDelaySeconds=10,periodSeconds=10,timeoutSeconds=1),
                      labels={@Label(key="runtimedb",value="mysql")},
                      ports={@Port(name="http",containerPort=8080),
                             @Port(name="https",containerPort=8443),
                             @Port(name="jolokia",containerPort=8778)
                            },
                     envVars={@Env(name="ENV_VAR",value="X_Y_Z")}
                    )
@S2iBuild(builderImage="registry.access.redhat.com/openjdk/openjdk-11-rhel7:latest",
          buildEnvVars={@Env(name="BUILD_EN_VAR",value="A_B_C")}
         )
@SpringBootApplication
public class RestApplication {

	public static void main(String[] args) {
		SpringApplication.run(RestApplication.class, args);
	}
}
```

cmoulliard:
- Add test case to reproduce the error. See lines 51-52 of the code
- Implement a first version of the new logic as defined within the ticket
- Ingress code is still not implemented as there is not Kubernetes Config property like `service.expose` to check if it has been enabled.

Fixes #775

### DIFF
--- a/core/src/main/java/io/dekorate/kubernetes/configurator/ApplyPort.java
+++ b/core/src/main/java/io/dekorate/kubernetes/configurator/ApplyPort.java
@@ -25,7 +25,6 @@ import io.dekorate.kubernetes.config.BaseConfigFluent;
 import io.dekorate.kubernetes.config.Configurator;
 import io.dekorate.kubernetes.config.Port;
 import io.dekorate.kubernetes.config.PortBuilder;
-import io.dekorate.utils.Ports;
 import io.dekorate.utils.Strings;
 
 public class ApplyPort extends Configurator<BaseConfigFluent<?>> {
@@ -63,11 +62,14 @@ public class ApplyPort extends Configurator<BaseConfigFluent<?>> {
         config.editMatchingPort(predicate)
             .withHostPort(port.getHostPort())
             .endPort();
-      } else if (Ports.isWebPort(port)) {
-        config.editMatchingPort(predicate)
-            .withHostPort(80)
-            .endPort();
-      }
+      } /*
+         * Delete to AddServiceResourceDecorator the role to define it
+         * else if (Ports.isWebPort(port)) {
+         * config.editMatchingPort(predicate)
+         * .withHostPort(80)
+         * .endPort();
+         * }
+         */
     } else {
       String name = names.size() > 0 ? names.get(0) : FALLBACK_PORT_NAME;
       config.addNewPortLike(port)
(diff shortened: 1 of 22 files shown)


iocanel (changes requested):
While the PR seems to address the issue, the solution is given at the decorator level, while I feel that the issue should be dealt at configuration level. 

Why?

Because we shouldn't allow invalid configuration data (containing dublicates) as they possibly affect multiple areas including:

- Ingress
- Probes
- Services
- Knative

and so on.
This is why we use two layers in the first place.
1. deal with cross cutting configuraton.
2. apply configuration to resources.

iocanel:
@cmoulliard: to speed things up I will pickup the pull request from here, if you don't mind.

iocanel:
Closing this one in favor of: #783

PR #783, opened 2021-08-31
Issue 775

iocanel:
Resolves: #775


core/src/main/java/io/dekorate/kubernetes/configurator/ApplyPort.java
```diff
@@ -34,45 +33,71 @@
...
+      matchFound = true;
+      applyPath(config, updated, matchingPortName);
+      applyContainerPort(config, updated, matchingPortName);
+
+    }
+
+    if (config.hasMatchingPort(matchingHostPort)) {
+      matchFound = true;
+      applyPath(config, updated, matchingPortName);
+      applyContainerPort(config, updated, matchingPortName);
```
aureamunoz, line 65: shouldn't be `applyContainerPort(config, updated, matchingHostPort);` instead?

iocanel: Great catch!

PR #785, opened 2021-09-01
fix: use hostPort predicate when mathcing host port is found in Apply…

iocanel:
…Port


iocanel:
@aureamunoz since you spotted my mistake please take a sec to review! Thanks!

aureamunoz (approved):
LGTM!

PR #800, opened 2021-11-03
Support injection of Routes and URLs

Sgitario:
This is useful when deploying apps that are exposed. This way we would not need to port forward the pods. 
Usage:
```java
@OpenshiftIntegrationTest
class SpringBootOnOpenshiftIT {
  @Inject
  private Route route;

  @RouteURL
  private URL appUrl;

  @Test
  public void shouldRespondWithHelloWorld() throws Exception {
    assertNotNull(route);
    assertNotNull(appUrl);

    OkHttpClient client = new OkHttpClient();
    Request request = new Request.Builder().get().url(appUrl).build();
    Response response = client.newCall(request).execute();
    assertEquals(response.body().string(), "Hello world");
  }

}
```


aureamunoz (approved):
nice. Good for me

PR #802, opened 2021-11-04
Fix broken master

iocanel:


examples/spring-boot-on-kubernetes-with-jib-custom-registry-example/src/main/resources/application.properties
```diff
@@ -1,2 +1,3 @@
 server.port=9090
 dekorate.jib.registry=quay.io
+dekorate.docker.enabled=false
```
aureamunoz, line 3: So, we need to disable explicitly conf other than the one we want to use?

aureamunoz (commented):
LGTM

examples/spring-boot-on-kubernetes-with-jib-custom-registry-example/src/main/resources/application.properties
iocanel: For now yes.
The next step is to prioritize build services that have been explicitly configured.
For example, if use configures jib and not docker jib gets priority.

aureamunoz: Got it!

aureamunoz (approved):

examples/spring-boot-on-kubernetes-with-jib-example/src/test/java/io/dekorate/example/SpringBootOnKubernetesTest.java
```diff
@@ -36,7 +36,6 @@ public void shouldContainDeployment() {
     Deployment d = findFirst(list, Deployment.class).orElseThrow(() -> new IllegalStateException());
     assertNotNull(d);
     Container container = d.getSpec().getTemplate().getSpec().getContainers().get(0);
-    assertTrue(container.getImage().contains("docker.io"));
```
Sgitario, line 39: This line was deleted on purpose? Thanks to this assertion this issue was spotted. cc @iocanel

PR #808, opened 2021-11-09
Testing: Don't delete resources if deploy was disabled

Sgitario:
When using `@KubernetesIntegrationTest(deployEnabled = false)` or `@OpenShiftIntegrationTest(deployEnabled = false)` means that the test framework did not deploy the resources, but they were deployed before running the test. 

The problem is that regardless if the testing deployed the application or not, it's always deleting the resources.

With these changes, the test framework will only delete the resources if and only if the test framework deployed them.


iocanel (approved):
LGTM

PR #812, opened 2021-11-10
Testing: Fix resolve of generated resources for Kubernetes

Sgitario:
Both Kubernetes and OpenShift resources should use the same logic to retrieve the resources to deploy.


iocanel (approved):
LGTM

PR #814, opened 2021-11-10
fix: avoid running itests that require docker.io access from forks

iocanel:


Sgitario (commented):
I would add a TODO comment in the disabled tests adding some context about why they are disabled and a link to an issue to enabled them back.

iocanel:
Todo comments added.

PR #815, opened 2021-11-11
Reenable tests for MultiModule projects

Sgitario:
And fix them if they are failing.


Sgitario:
Accidental merged. Already fixed in Master.

PR #817, opened 2021-11-11
Fix Diagnosis for failures ++ Update Disabled Tests with linking issue

Sgitario:
Update the disabled message for failures with the following issue: https://github.com/dekorateio/dekorate/issues/818


examples/spring-boot-on-kubernetes-example/pom.xml
```diff
@@ -30,12 +30,6 @@
       <version>${version.spring-boot}</version>
     </dependency>
 
-    <dependency>
```
iocanel, line 33: We need this.

Sgitario: Please, don't review this PR yet. It's a try and error PR, to investigate what is the root cause of the failures.

Sgitario: Update PR to fix the diagnosis only.

iocanel (approved):

Issue #821, opened 2021-11-22
Testing: Docker build is always run even if it's turned off when there is Dockerfile present


Sgitario:
Using `@OpenshiftIntegrationTest` and disabling the docker build:
```
dekorate.docker.enabled=false
dekorate.s2i.enabled=true
```

The Docker build is being triggered:
```
[INFO] Performing docker build.
```

This is happening either because the Dockerfile is present in the project or because we're also adding the kubernetes junit5 dependency.

aureamunoz:
I had notified something similar using JIB

PR #822, opened 2021-11-22
Testing: Docker build should not be used if it's disabled

Sgitario:
The property `dekorate.docker.enabled` was being ignored by the test framework.
It also fixes the `jib` build.

Fix https://github.com/dekorateio/dekorate/issues/821


annotations/s2i-annotations/src/main/java/io/dekorate/s2i/manifest/S2iManifestGenerator.java
```diff
@@ -91,9 +91,8 @@ public void generate(S2iBuildConfig config) {
       ImageConfiguration imageConfig = configurationRegistry
           .getImageConfig(BuildServiceFactories.supplierMatches(getProject()))
           .orElse(null);
-      String registry = Strings.isNullOrEmpty(imageConfig.getRegistry()) ? "docker.io" : imageConfig.getRegistry();
-
       if (imageConfig != null) {
+        String registry = Strings.isNullOrEmpty(imageConfig.getRegistry()) ? "docker.io" : imageConfig.getRegistry();
```
Sgitario, line 95: This is to avoid null pointer exception which happens when we disable the s2i build.

iocanel (approved):
LGTM

PR #824, opened 2021-12-02
Support configuration based on spring/dekorate profiles, and components

Sgitario:
New order of properties file to use:

1. Annotations
2. `application.properties`
3. `application.yaml`
4. `application.yml`

Then,  it will use the properties file depending on the active Dekorate dependencies in use. For example, if we're using the dependency `io.dekorate:kubernetes-annotations`, then:
1. `application-kubernetes.properties`
2. `application-kubernetes.yaml`
3. `application-kubernetes.yml`

| Note that only the `openshift`, `kubernetes` and `knative` modules are providing additional properties files.

Then, for Spring Boot applications, it will also take into account the Spring property `spring.profiles.active` if set:
1. `application-${spring.profiles.active}.properties`
2. `application-${spring.profiles.active}.yaml`
3. `application-${spring.profiles.active}.yml`

Finally, if the Dekorate profile property `dekorate.properties.profile` is set:
1. if property `dekorate.properties.profile` is set, then `application-${dekorate.properties.profile}.properties`
2. if property `dekorate.properties.profile` is set, then `application-${dekorate.properties.profile}.yaml`
3. if property `dekorate.properties.profile` is set, then `application-${dekorate.properties.profile}.yml`

### DIFF
--- a/annotations/knative-annotations/src/main/java/io/dekorate/knative/config/KnativeAdditionalResourcesProvider.java
+++ b/annotations/knative-annotations/src/main/java/io/dekorate/knative/config/KnativeAdditionalResourcesProvider.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2018 The original authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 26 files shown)


iocanel (approved):
LGTM. There are some test failures that need to be addressed first.

Sgitario:
> LGTM. There are some test failures that need to be addressed first.

Fixed!

PR #828, opened 2021-12-16
feat: add kind support for loading docker images

### LINKED ISSUE #810: Add support for kind
It would be great if we had a to automatically do `kind load docker-image` for built images.
When this feature is enabled it should trigger the `load` right after the build. (
It should also work from within the testing framework (e.g. when building an image to automatically load it to kind).

This would also add great value to the testing framework.

# Implementation ideas
We could either create a new module or add support for Docker and Jib builds (for jib it doesn't make sense for all use cases, but only when a local docker daemon is used).

aureamunoz:
Fixes #810

New module for kind support


iocanel (changes requested):
I think that there are more in things in this pull request than we actually need.

For example, one could completely drop the `kind-annotations` module and things would still work by specifying something like `dekorate.docker.auto-load=true` (possibly with a few quirks here and there).

The same might be possibly true the other way around.
One could completely drop the `core` changes and keep `kind-annotations` and be really close to get it working (the only issue we would have is ensuring that the auto-load thing would trigger the right moment.

I tend to prefer ditching the `core` changes and keeping `kind-annotations` as a module but I am not 100% sure about this. Let's schedule a meeting to further discuss it.

aureamunoz:
Ok, I had done this way in the firs attemp (I have a branch with it) but then I thought that I worth to do it in a dedicated module...I will push it tomorrow

aureamunoz (commented):
That's perfect. Thanks!!

PR #847, opened 2022-01-21
Minikube module

aureamunoz:
Here is the PR to support minikube.
I have added a new module which generates a `minikube.yaml` manifest with a NodePort Service with some default values. At the moment the PR is lacking the part to gather the ports from config or range. 
As seen together the MinikubeManifestGenerator is the same that KuberentesManifestGenerator.
I did some refactor also to align some classes packages in the Spring framework part with other modules, do you remember the change of Handler names to ManifestGenerator, they were also move into `config` packages so I did it for Spring.


aureamunoz:
I think you can already review this @iocanel, let's see what CI says

aureamunoz:
I will try to fix the sonar checks in another branch, they are due to the new site code, not the code related to this PR

aureamunoz:
> I will try to fix the sonar checks in another branch, they are due to the new site code, not the code related to this PR

Fixed in the last commit

iocanel (commented):
I leaning towards having the minikube annotation be something that we will use side by side with @KubernetesApplication (e.g. like we do with @Kind).

I will provide detailed feedback shortly.

annotations/minikube-annotations/src/main/java/io/dekorate/minikube/config/DefaultMinikubeConfigGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+import io.dekorate.project.ApplyProjectInfo;
+
+public class DefaultMinikubeConfigGenerator implements MinikubeConfigGenerator {
+
+  private final ConfigurationRegistry configurationRegistry;
+
+  public DefaultMinikubeConfigGenerator(ConfigurationRegistry configurationRegistry) {
+    this.configurationRegistry = configurationRegistry;
+    this.configurationRegistry.add(new ApplyProjectInfo(getProject()));
+    this.configurationRegistry.add(new ApplyServiceTypeNodePortConfigurator());
```
iocanel, line 31: `Configurators` modify the configuration objects that is passed around to the rest of system.
As a result, these changes will be passed to Kubernetes, Openshift etc, which is not desirable.
We only want these changes to be applied for minikube (i.e. minikube.yml/json).

core/src/main/java/io/dekorate/utils/Ports.java
```diff
@@ -75,15 +82,15 @@
   }
 
   public static Port populateHostPort(Port port) {
-    if (!isWebPort(port)) {
+    if (!isWebPort(port) && !isNodePort(port)) {
```
iocanel, line 85: It's not obvious to me why were are also checking `isNodePort`. A comment might help.

annotations/minikube-annotations/src/main/java/io/dekorate/minikube/annotation/MinikubeApplication.java
```diff
@@ -0,0 +1,270 @@
...
+import io.fabric8.kubernetes.api.model.Service;
+import io.sundr.builder.annotations.Adapter;
+import io.sundr.builder.annotations.Buildable;
+import io.sundr.builder.annotations.Pojo;
+
+@Buildable(builderPackage = "io.fabric8.kubernetes.api.builder")
+@Pojo(name = "MinikubeConfig", relativePath = "../config", autobox = true, mutable = true, superClass = BaseConfig.class, withStaticBuilderMethod = true, withStaticAdapterMethod = false, adapter = @Adapter(suffix = "Adapter", relativePath = "../adapter", withMapAdapterMethod = true))
+@Target({ ElementType.CONSTRUCTOR, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MinikubeApplication {
```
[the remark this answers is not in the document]
aureamunoz: done

core/src/main/java/io/dekorate/utils/Ports.java
aureamunoz: I removed this check in the last commit, I think that a node port service is similar to a cluster IP one, a connection received on the port will be forwarded to a pod so we need to populate the host port also in that case. This is also the reason why the `PopoluateWebPort` configurator [is added](https://github.com/aureamunoz/dekorate/blob/0831626fb58707f0eca9f1d9bcdc7e839f04dcbe/annotations/minikube-annotations/src/main/java/io/dekorate/minikube/config/DefaultMinikubeConfigGenerator.java#L32) in the Minikube module

aureamunoz (commented):
:+1: 
We need now to add this to the docs, I will create an issue to do it in other PR, wdyt @iocanel ?

iocanel:
> +1 We need now to add this to the docs, I will create an issue to do it in other PR, wdyt @iocanel ?

Eitherway is fine by me.

iocanel (approved):
LGTM

annotations/minikube-annotations/src/main/java/io/dekorate/minikube/annotation/Minikube.java
```diff
@@ -0,0 +1,57 @@
...
+@Target({ ElementType.CONSTRUCTOR, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Minikube {
+
+  /**
+   * The number of replicas to use.
+   *
+   * @return The number of replicas.
+   */
+  int replicas() default 1;
```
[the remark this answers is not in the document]
aureamunoz: yes, it's removed, this is an outdated version of the file, don't know why

aureamunoz:
I finally added docs for minikube and rebased main branch. Could you please take a look? @iocanel

PR #848, opened 2022-01-27
Fix JSON auto generated files

Sgitario:
At the moment, the auto-generated file `kubernetes.json` looks like:

```json
{ object 1}{ object 2}{ object 3}
```

Which is not well formatted. This PR fixes this and produces the JSON file in the right format:

```json
[{ object 1}, 
{ object 2},
{ object 3}]
```


iocanel (approved):
LGTM

PR #850, opened 2022-01-28
Fix path resolution: User configuration will now take precedence

Sgitario:
Before these changes, if users have the following controller:

```java
@RestController
@RequestMapping("/api")
public class HelloController {

  private static final String HELLO = "hello world!";

  @RequestMapping("/")
  public String hello() {
    return HELLO;
  }
}
```

Dekorate generates a route (or ingress) with the path `/api`:

```
---
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  // ...
  name: issue-545-spring-boot-web-with-port-path-openshift
spec:
  path: /api
  port:
    targetPort: 8080
// ...
```

The problem is that if we want to overwrite the `path` field, the only way is to use the properties:

```
dekorate.openshift.expose=true
dekorate.openshift.ports[0].name=http
dekorate.openshift.ports[0].containerPort=8080
dekorate.openshift.ports[0].path=/app
```

But these properties were not working because the calculated properties were also taken priority over the user properties. 

This PR addresses this issue to always use the user properties if provided.


iocanel (changes requested):
I am a bit reluctant on changing the behavior of the existing `Configurator`, as this might have impact that our tests don't catch or even break downstream projects. I would prefer if we created an alternative `Configurator` or if we somehow made sure that this is not a breaking change.

iocanel:
> > I am a bit reluctant on changing the behavior of the existing `Configurator`, as this might have impact that our tests don't catch or even break downstream projects. I would prefer if we created an alternative `Configurator` or if we somehow made sure that this is not a breaking change.
> 
> I don't think the solution should be to bring a new `Configurator` as the existing `Configurator` is not working well at the moment. Actually, the existing tests that are meant to cover this are wrong. If you amend the wrong piece of code (https://github.com/dekorateio/dekorate/pull/850/files#diff-3dae82698228d378b7cd00b88b9e44b57ccab7dbcaf8c589ca55dd0f4a02fde0R24), the test will stop working.
> 
> Also, I don't see what scenarios can potentially be broken with this PR. Before this PR, the port `path` properties from the user were always ignored. Still, my experience with Dekorate is very little, so let me know if what you're asking is to cover some scenario I'm not seeing and I will be happy to add it, better safe than sorry!

I son

> > I am a bit reluctant on changing the behavior of the existing `Configurator`, as this might have impact that our tests don't catch or even break downstream projects. I would prefer if we created an alternative `Configurator` or if we somehow made sure that this is not a breaking change.
> 
> I don't think the solution should be to bring a new `Configurator` as the existing `Configurator` is not working well at the moment. Actually, the existing tests that are meant to cover this are wrong. If you amend the wrong piece of code (https://github.com/dekorateio/dekorate/pull/850/files#diff-3dae82698228d378b7cd00b88b9e44b57ccab7dbcaf8c589ca55dd0f4a02fde0R24), the test will stop working.
> 
> Also, I don't see what scenarios can potentially be broken with this PR. Before this PR, the port `path` properties from the user were always ignored. Still, my experience with Dekorate is very little, so let me know if what you're asking is to cover some scenario I'm not seeing and I will be happy to add it, better safe than sorry!

The sceanrio that is not covered is two different parts of dekorate contributing different attributes of a port configuration.
**Example**: A spring app that uses a custom `server.port` and the user configuring a custom path. This will not work  with the existing configurator. Your test works, because user configuration overrides both the port number and path.

Sgitario:
> > > I am a bit reluctant on changing the behavior of the existing `Configurator`, as this might have impact that our tests don't catch or even break downstream projects. I would prefer if we created an alternative `Configurator` or if we somehow made sure that this is not a breaking change.
> > 
> > 
> > I don't think the solution should be to bring a new `Configurator` as the existing `Configurator` is not working well at the moment. Actually, the existing tests that are meant to cover this are wrong. If you amend the wrong piece of code (https://github.com/dekorateio/dekorate/pull/850/files#diff-3dae82698228d378b7cd00b88b9e44b57ccab7dbcaf8c589ca55dd0f4a02fde0R24), the test will stop working.
> > Also, I don't see what scenarios can potentially be broken with this PR. Before this PR, the port `path` properties from the user were always ignored. Still, my experience with Dekorate is very little, so let me know if what you're asking is to cover some scenario I'm not seeing and I will be happy to add it, better safe than sorry!
> 
> I son
> 
> > > I am a bit reluctant on changing the behavior of the existing `Configurator`, as this might have impact that our tests don't catch or even break downstream projects. I would prefer if we created an alternative `Configurator` or if we somehow made sure that this is not a breaking change.
> > 
> > 
> > I don't think the solution should be to bring a new `Configurator` as the existing `Configurator` is not working well at the moment. Actually, the existing tests that are meant to cover this are wrong. If you amend the wrong piece of code (https://github.com/dekorateio/dekorate/pull/850/files#diff-3dae82698228d378b7cd00b88b9e44b57ccab7dbcaf8c589ca55dd0f4a02fde0R24), the test will stop working.
> > Also, I don't see what scenarios can potentially be broken with this PR. Before this PR, the port `path` properties from the user were always ignored. Still, my experience with Dekorate is very little, so let me know if what you're asking is to cover some scenario I'm not seeing and I will be happy to add it, better safe than sorry!
> 
> The sceanrio that is not covered is two different parts of dekorate contributing different attributes of a port configuration. **Example**: A spring app that uses a custom `server.port` and the user configuring a custom path. This will not work with the existing configurator. Your test works, because user configuration overrides both the port number and path.

So, the scenario is users setting `server.port` and having the port path configuration:

```
server.port=5000

dekorate.kubernetes.expose=true
dekorate.kubernetes.ports[0].name=http
dekorate.kubernetes.ports[0].path=/app
```

This scenario will set the path `/app` to the port 5000 as expected. So, it will be working fine.

I guess the problem you foresee is when users also set the `containerPort`:

```
server.port=5000

dekorate.kubernetes.expose=true
dekorate.kubernetes.ports[0].name=http
dekorate.kubernetes.ports[0].containerPort=8080
dekorate.kubernetes.ports[0].path=/app
```

In this scenario, the port in `server.port` is used in the container spec, but the port in the service will be 8080. But as far I see it, this is a wrong user configuration, not a wrong behaviour of the Configurator :/

iocanel:
> In this scenario, the port in `server.port` is used in the container spec, but the port in the service will be 8080. But as far I see it, this is a wrong user configuration, not a wrong behaviour of the Configurator :/

While I agree that in this case user provides contradicting configuration, I feel that we should still produce a predictable result. In other words the dekorate user config should take precedence over the framework config.

Sgitario:
> > In this scenario, the port in `server.port` is used in the container spec, but the port in the service will be 8080. But as far I see it, this is a wrong user configuration, not a wrong behaviour of the Configurator :/
> 
> While I agree that in this case user provides contradicting configuration, I feel that we should still produce a predictable result. In other words the dekorate user config should take precedence over the framework config.

+100
In this case, the user config is providing both `server.port` and `dekorate.kubernetes.ports[0].containerPort` with a different value. From my point of view, the most that Dekorate could do here It's to fail by saing: "`http` was configured by the user with `8080` and now you're using `5000`!". However, I'm not sure if there is an easy way to detect that something is coming from the user configuration or from the defaults, so not sure the effort to implement this is worthy.

iocanel:
> > > In this scenario, the port in `server.port` is used in the container spec, but the port in the service will be 8080. But as far I see it, this is a wrong user configuration, not a wrong behaviour of the Configurator :/
> > 
> > 
> > While I agree that in this case user provides contradicting configuration, I feel that we should still produce a predictable result. In other words the dekorate user config should take precedence over the framework config.
> 
> +100 In this case, the user config is providing both `server.port` and `dekorate.kubernetes.ports[0].containerPort` with a different value. From my point of view, the most that Dekorate could do here It's to fail by saing: "`http` was configured by the user with `8080` and now you're using `5000`!". However, I'm not sure if there is an easy way to detect that something is coming from the user configuration or from the defaults, so not sure the effort to implement this is worthy.

I think that we are overthinking it. The problem we are trying to solve is the following: `do not let framework config take precedence over dekorate config`. This is as simple as adding a flag to the configurator or creating a new one that never modifies existing  ports. This is pretty much the advertised behavior, I don't feel that at this point we need to risk drastically changing the behvior.

Sgitario:
> > > > In this scenario, the port in `server.port` is used in the container spec, but the port in the service will be 8080. But as far I see it, this is a wrong user configuration, not a wrong behaviour of the Configurator :/
> > > 
> > > 
> > > While I agree that in this case user provides contradicting configuration, I feel that we should still produce a predictable result. In other words the dekorate user config should take precedence over the framework config.
> > 
> > 
> > +100 In this case, the user config is providing both `server.port` and `dekorate.kubernetes.ports[0].containerPort` with a different value. From my point of view, the most that Dekorate could do here It's to fail by saing: "`http` was configured by the user with `8080` and now you're using `5000`!". However, I'm not sure if there is an easy way to detect that something is coming from the user configuration or from the defaults, so not sure the effort to implement this is worthy.
> 
> I think that we are overthinking it. The problem we are trying to solve is the following: `do not let framework config take precedence over dekorate config`. This is as simple as adding a flag to the configurator or creating a new one that never modifies existing ports. This is pretty much the advertised behavior, I don't feel that at this point we need to risk drastically changing the behvior.

Indeed, this is what the changes in this PR do: `do not let framework config take precedence over dekorate config`. 

Don't follow you what you stand for adding a flag to the current configurator, but unless I'm missing something I disagree about creating a new configurator because it would not fix the existing configurator which is overwriting the user configuration.

Sgitario:
> I've just found an issue with the changes in this PR. When using the following user configuration:
> 
> ```
> server.port=5000
> 
> dekorate.kubernetes.expose=true
> dekorate.kubernetes.ports[0].name=http
> dekorate.kubernetes.ports[0].containerPort=5000
> dekorate.kubernetes.ports[0].path=/app
> ```
> 
> The property `dekorate.kubernetes.ports[0].containerPort` is redundant, however it causes the ports being duplicated in the `target/.dekorate/config/kubernetes.yml` object:
> 
> ```
> host: null
> ports:
>   - name: http
>     containerPort: 5000
>     hostPort: 80
>     path: /app
>     protocol: TCP
>   - name: http
>     containerPort: 5000
>     hostPort: 80
>     path: /api
>     protocol: null
> ```
> 
> Where we should only expect one port:
> 
> ```
> host: null
> ports:
>   - name: http
>     containerPort: 5000
>     hostPort: 80
>     path: /app
>     protocol: null
> ```
> 
> This resource is used by the test framework and may potentially have some other side effects. Therefore, it needs to be fixed in this PR.
> 
> fyi @iocanel

Fixed by changing the condition to not add the calculated port unless there is one matching port from user configuration by name and host name. https://github.com/dekorateio/dekorate/pull/850/files#diff-916fe474c30a297209cff78749e2b914536ce8efbfaf854f20c734bd050dcb36R48

iocanel (approved):
LGTM

PR #851, opened 2022-02-02
Spring Boot: Change default actuator liveness/readiness probes

Sgitario:
From Spring Boot 2.3, the `/actuator/info` endpoint which is used by the liveness probe, will only be exposed if and only if there is data to present (or it's forced by using `management.endpoints.web.exposure.include=info`).

Moreover, from Spring Boot 2.3, we should use the health liveness / readiness endpoints.


aureamunoz (approved):
These are breaking changes, shouldn't we indicate it somewhere? or add something about it in the docs

Sgitario:
> These are breaking changes, shouldn't we indicate it somewhere? or add something about it in the docs

+1. There is no mention to this in the docs, that's why I didn't update anything there. But let me know how to notify the breaking changes in the release notes.

iocanel (approved):
LGTM

PR #852, opened 2022-02-03
Fix parsing in `Git.getRemotes` to ignore submodules

kdubb:
The parsing of git config in the `Git.getRemotes` now properly ignores submodule references.

Previously, a any submodule sections would add a null keyed entry to the hash map. Aside from _not_ being a remote, it causes `DekorateException`s on output with the message "JSON doesn't allow `null` as a map key".

### DIFF
--- a/core/src/main/java/io/dekorate/utils/Git.java
+++ b/core/src/main/java/io/dekorate/utils/Git.java
@@ -23,10 +23,10 @@ package io.dekorate.utils;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


kdubb:
@Sgitario This is a fairly serious issue as it causes build crashes in Quarkus for any project that uses submodules with `quarkus-kubernetes`.

Fast-tracking this and a quick release would be very helpful.

geoand (approved):
LGTM!

geoand:
> Fast-tracking this and a quick release would be very helpful.

I defintely agree with this. @iocanel we you make that happen please?

iocanel:
CI failures seem to be intermittent. Let me re-run them and most probably I'll merge anyway. I intend to release that tomorrow.

kdubb:
@iocanel Looks like we're good!

geoand:
@iocanel is there a release that contains this fix?

Asking because we would need it in Quarkus by tonight in order for the fix to make it into `2.7.1.Final`

PR #853, opened 2022-02-03
fix: getDeploymentMetadata now filters by name

iocanel:


annotations/s2i-annotations/src/main/java/io/dekorate/s2i/decorator/AddBuildConfigResourceDecorator.java
```diff
@@ -51,7 +51,9 @@ public AddBuildConfigResourceDecorator(S2iBuildConfig config) {
   }
 
   public void visit(KubernetesListBuilder list) {
-    ObjectMeta meta = getMandatoryDeploymentMetadata(list);
+    List<ObjectMeta> metaList = getDeploymentMetadataList(list, this.config.getName());
+    //If there are multiple deployment resource we can't possibly determine which one to use.
+    ObjectMeta meta = metaList.size() == 1 ? metaList.get(0) : new ObjectMeta();
```
Sgitario, line 56: These three lines are equivalent to simply do:
`getDeploymentMetadata(list, this.config.getName()).orElseGet(ObjectMeta::new);`

annotations/s2i-annotations/src/main/java/io/dekorate/s2i/decorator/AddBuilderImageStreamResourceDecorator.java
```diff
@@ -43,8 +43,9 @@ public AddBuilderImageStreamResourceDecorator(S2iBuildConfig config) {
   }
 
   public void visit(KubernetesListBuilder list) {
-    ObjectMeta meta = getMandatoryDeploymentMetadata(list);
-
+    List<ObjectMeta> metaList = getDeploymentMetadataList(list, this.config.getName());
+    //If there are multiple deployment resource we can't possibly determine which one to use.
+    ObjectMeta meta = metaList.size() == 1 ? metaList.get(0) : new ObjectMeta();
```
Sgitario, line 48: These three lines are equivalent to simply do:
`getDeploymentMetadata(list, this.config.getName()).orElseGet(ObjectMeta::new);`

annotations/s2i-annotations/src/main/java/io/dekorate/s2i/decorator/AddDockerImageStreamResourceDecorator.java
```diff
@@ -36,7 +38,9 @@ public AddDockerImageStreamResourceDecorator(ImageConfiguration config, String d
   }
 
   public void visit(KubernetesListBuilder list) {
-    ObjectMeta meta = getMandatoryDeploymentMetadata(list);
+    List<ObjectMeta> metaList = getDeploymentMetadataList(list, this.config.getName());
+    //If there are multiple deployment resource we can't possibly determine which one to use.
+    ObjectMeta meta = metaList.size() == 1 ? metaList.get(0) : new ObjectMeta();
```
Sgitario, line 43: These three lines are equivalent to simply do:
`getDeploymentMetadata(list, this.config.getName()).orElseGet(ObjectMeta::new);`

annotations/s2i-annotations/src/main/java/io/dekorate/s2i/decorator/AddOutputImageStreamResourceDecorator.java
```diff
@@ -34,7 +36,9 @@ public AddOutputImageStreamResourceDecorator(S2iBuildConfig config) {
   }
 
   public void visit(KubernetesListBuilder list) {
-    ObjectMeta meta = getMandatoryDeploymentMetadata(list);
+    List<ObjectMeta> metaList = getDeploymentMetadataList(list, this.config.getName());
+    //If there are multiple deployment resource we can't possibly determine which one to use.
+    ObjectMeta meta = metaList.size() == 1 ? metaList.get(0) : new ObjectMeta();
```
Sgitario, line 41: These three lines are equivalent to simply do:
`getDeploymentMetadata(list, this.config.getName()).orElseGet(ObjectMeta::new);`

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/manifest/TektonManifestGenerator.java
```diff
@@ -389,6 +385,10 @@ public void generatePipelineResources(TektonConfig config) {
       resourceRegistry.decorate(TEKTON_PIPELINE_RUN,
           new AddPvcToPipelineRunDecorator(null, PIPELINE_M2_WS, m2WorkspaceClaimName, false));
     }
+
+    resourceRegistry.decorate(TEKTON_PIPELINE, new AddServiceAccountResourceDecorator(imageBuildTaskName));
```
Sgitario, line 389: I think this is not correct. Before, this decorator was using `config.getName()` and then it will retrieve the `Deployment` (or Service, DeploymentConfig, ...) resources by this name (see line https://github.com/dekorateio/dekorate/pull/853/files#diff-11fbc30c8c05acda88e00bd0e05f40d2dd1fda355ca6b672e2918a1354c78b05R37).

However, this decorator now uses `imageBuildTaskName`, so the Deployment resources will never be found.

core/src/main/java/io/dekorate/kubernetes/decorator/AddRoleBindingResourceDecorator.java
```diff
@@ -54,7 +54,7 @@ public AddRoleBindingResourceDecorator(String name, String serviceAccount, Strin
   }
 
   public void visit(KubernetesListBuilder list) {
-    ObjectMeta meta = getMandatoryDeploymentMetadata(list);
+    ObjectMeta meta = getMandatoryDeploymentMetadata(list, this.name);
```
Sgitario, line 57: Is this being covered by some tests? Looking at https://github.com/dekorateio/dekorate/pull/853/files#diff-d63a9fd5811d5b671a65f15670a2f2d90a61d424fb4f17073a649cdffad5f93fR390, I think this is never going to find the metadata by the name because it's using something like "name + :deployer", but maybe I'm missing something.

core/src/main/java/io/dekorate/kubernetes/decorator/AddDockerConfigJsonSecretDecorator.java
```diff
@@ -67,7 +67,7 @@ public AddDockerConfigJsonSecretDecorator(String name, Path existingConfigJson,
 
   @Override
   public void visit(KubernetesListBuilder list) {
-    String name = Strings.isNotNullOrEmpty(this.name) ? this.name : getMandatoryDeploymentMetadata(list).getName();
+    String name = Strings.isNotNullOrEmpty(this.name) ? this.name : getMandatoryDeploymentMetadata(list, ANY).getName();
```
Sgitario, line 70: I would read this line easier with `getDeploymentMetadata(list, ANY).map(ObjectMeta::getName).orElse(name)`

core/src/main/java/io/dekorate/kubernetes/decorator/ResourceProvidingDecorator.java
```diff
@@ -51,22 +51,32 @@ public boolean match(HasMetadata h, String apiVersion, String kind, String name)
     return true;
   }
 
-  public Optional<ObjectMeta> getDeploymentMetadata(KubernetesListBuilder list) {
-    return list.getItems().stream().filter(h -> DEPLOYMENT_KINDS.contains(h.getKind())).map(HasMetadata::getMetadata)
-        .findFirst();
+  public List<ObjectMeta> getDeploymentMetadataList(KubernetesListBuilder list, String name) {
+    String trimedName = Strings.isNotNullOrEmpty(name) ? name.replaceAll("[:].*$", "") : name;
```
Sgitario, line 55: I don't see the purpose of this replacement `name.replaceAll("[:].*$", "")`. I think it's new in this PR. I think adding a comment would be helpful to understand this rule.

Sgitario (changes requested):

aureamunoz (approved):
I aggree with Jose that a comment in the regex would be useful. Apart from that LGTM

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/manifest/TektonManifestGenerator.java
iocanel: Before, this argument was just used to name the resource and didn't affect discovery of existing resources.
Now, that discovery by name is introduced we need to make sure that we pass the correct name, in this case `imageBuildTaskName`.

The discovery part works fine, according to the tests, but you may be right that this change in naming may introduce hiccups. I'll have check to see if we have issues there.

core/src/main/java/io/dekorate/kubernetes/decorator/AddRoleBindingResourceDecorator.java
iocanel: What you are missing is that the suffix is `stripped`

annotations/s2i-annotations/src/main/java/io/dekorate/s2i/decorator/AddDockerImageStreamResourceDecorator.java
Sgitario: This comment was marked as resolved, but it was not updated. Did you do it on purpose?

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/manifest/TektonManifestGenerator.java
Sgitario: +1

core/src/main/java/io/dekorate/kubernetes/decorator/ResourceProvidingDecorator.java
```diff
@@ -51,22 +51,35 @@ public boolean match(HasMetadata h, String apiVersion, String kind, String name)
     return true;
   }
 
-  public Optional<ObjectMeta> getDeploymentMetadata(KubernetesListBuilder list) {
-    return list.getItems().stream().filter(h -> DEPLOYMENT_KINDS.contains(h.getKind())).map(HasMetadata::getMetadata)
-        .findFirst();
+  public List<ObjectMeta> getDeploymentMetadataList(KubernetesListBuilder list, String name) {
+    // In 99% of the cases we select metadata by name.
+    // There are some edge cases (e.g. RoleBindings) where a suffix is added (e.g. <name>:deployer).
+    // We need to get rid of such suffixes when present as we NEVER have thtem in `deployment` kinds.
```
Sgitario, line 57:
```suggestion
    // We need to get rid of such suffixes when present as we NEVER have them in `deployment` kinds.
```

Sgitario (approved):

PR #859, opened 2022-02-07
refactor: fix website job

### LINKED ISSUE #860: Ensure dekorate site points to the new version each time we relese
For our releases we do have a script that among others updates the versions in the readme (see [1]). This scripts is actually calling via jbang a java program that does the job (see [2]).

We need to make sure that these scripts also update the files that we use to populate the site, as currently the site is displaying the wrong version.

[1]: https://github.com/dekorateio/dekorate/blob/main/release.sh#L19
[2]: https://github.com/dekorateio/dekorate/blob/main/scripts/ChangeVersion.java

aureamunoz:
The github action triggers on a tag push. It will set the correct version in the [file](https://github.com/dekorateio/dekorate/blob/9d56563ec00d087e1e055a6a690f36deb6f3db90/docs/_data/project.yml#L4-L3) used by Jekyll for gathering the Dekorate version in the whole site.

Fixes #860

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -4,18 +4,24 @@ on:
   push:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+*'
+
 jobs:
   publish-website:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set env
-        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      - name: Render docs and publish
+      - name: Set version
+        run: |
+          sudo apt-get install moreutils
+          .github/set-version.sh $GITHUB_REF_NAME
+      - name: Commit files
         run: |
           git config --global user.name "Dekorate"
           git config --global user.email "cmoullia-staff@redhat.com"
-          sudo apt-get install moreutils
-          echo $RELEASE_VERSION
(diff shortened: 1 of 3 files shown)


.github/workflows/website.yml
```diff
@@ -12,10 +15,16 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_API_TOKEN }}
         run: |
           git config --global user.name "Dekorate"
           git config --global user.email "cmoullia-staff@redhat.com"
           sudo apt-get install moreutils
           echo $RELEASE_VERSION
           echo ${{ env.RELEASE_VERSION }}
+          echo ${GITHUB_TOKEN}
+          echo 'about to clone repo'
+          git clone https://aureamunoz:${GITHUB_TOKEN}@github.com/dekorateio/dekorate.git
```
iocanel, line 28: I feel that there should be a way to deal with it without having to use our actual user names, email addresses etc.

Using `actions/checkout` could possibly be a more louse approach to it. If there are technical reasons why we need to add a comment as its not obvious.

iocanel (changes requested):

.github/workflows/website.yml
aureamunoz: yes, sure! I'm just debugging/testing....I should have created the PR in draft mode, I will

aureamunoz:
I will update Dekorate version in the current site manually and let's see if the job works when the next release

PR #864, opened 2022-02-14
Helm: Change file writer to session listener ++ Optionally create tar file ++ Add alias in dependency

Sgitario:
This PR would help to use the Helm extension externally.
In addition, it will add two additional fields to: optionally produce the Helm tar file ++ add the ` alias` property in Helm dependencies.


iocanel:
@Sgitario: Using `++` in the PR title is a strong indication that it might be a good idea to split the PR into multiple PRs. No need to do it now, but it's something you should probably consider in the future, as it makes it so much simpler for the reviewer.

iocanel (approved):
Overall LGTM.

It's not 100% clear to me, why you are converting writer to session listener (but I have some ideas). Other than that the changes seem reasonable.

Sgitario:
> Overall LGTM.
> 
> It's not 100% clear to me, why you are converting writer to session listener (but I have some ideas). Other than that the changes seem reasonable.

Sorry about not claryfing why this is needed.
I guess you figure it out tho. When using Dekorate in Quarkus, the file writer is being overwritten, so it's not being used at all. I tried with session listener and worked fine.

Issue #869, opened 2022-02-21
Wrong generated resources for Knative


Sgitario:
The Knative generated manifests have the following issues:
- The generated `config-autoscaler` configmap does not set the namespace which should always be `knative-serving` (see for example https://knative.dev/docs/serving/autoscaling/scale-to-zero/#enable-scale-to-zero).
- Some properties do add the annotations at Knative service metadata level, and it should add them at Knative service spec template metadata level.
- The property `dekorate.knative.global-auto-scaling.containerConcurrency` (hard limit option) wrongly uses the `config-autoscaler` configmap and it should use `config-defaults` configmap (see https://knative.dev/docs/serving/autoscaling/concurrency/#hard-limit)

PR #870, opened 2022-02-21
Knative: Fix generated resources

Sgitario:
The Knative generated manifests have the following issues:
- The generated `config-autoscaler` configmap does not set the namespace which should always be `knative-serving` (see for example https://knative.dev/docs/serving/autoscaling/scale-to-zero/#enable-scale-to-zero).
- Some properties do add the annotations at Knative service metadata level, and it should add them at Knative service spec template metadata level.
- The property `dekorate.knative.global-auto-scaling.containerConcurrency` (hard limit option) wrongly uses the `config-autoscaler` configmap and it should use `config-defaults` configmap (see https://knative.dev/docs/serving/autoscaling/concurrency/#hard-limit)

Fix https://github.com/dekorateio/dekorate/issues/869


iocanel (approved):
LGTM

PR #874, opened 2022-02-23
Fix tests in frameworkless example

### LINKED ISSUE #865: Openshift tests have become flaky
It seems that the last couple of weeks our openshift tests have gone flaky and we have intermitent failures.

All failures seems to be quite similar. We are trying to access the application before the endpoint is ready. 
So, my guess is that for some reason the probes are not setup correctly.

The weird issue is that this only seems to affect Openshift.

ammbra:
This PR should fix #865. 

You can observe a successful run of the GH Actions here: https://github.com/ammbra/dekorate/actions/runs/1887904057

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -20,7 +20,7 @@ env:
 on:
   push:
     branches:
-      - fix-test-frameworkless-example
+      - main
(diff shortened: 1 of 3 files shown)


iocanel (approved):
LGTM

Issue #854, opened 2022-02-04
Support Startup Probes for Kubernetes, Openshift and Knative


Sgitario:
# Description
The kubelet uses startup probes to know when a container application has started. If such a probe is configured, it disables liveness and readiness checks until it succeeds, making sure those probes don't interfere with the application startup. This can be used to adopt liveness checks on slow starting containers, avoiding them getting killed by the kubelet before they are up and running.

```yaml
startupProbe:
  httpGet:
    path: /healthz
    port: liveness-port
  failureThreshold: 30
  periodSeconds: 10
```

More context in [kubernetes.io/startup-probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).

# User story

```
Given an user who configures a startup probe via configuration
When dekorate generates the kubernetes or openshift or Knative or other relevant manifests
Then startup probes should be generated
```

```
Given an user who does NOT configure any startup probe
When dekorate generates the kubernetes or openshift or other relevant manifests
Then startup probes should NOT be generated
```

# Acceptance criteria
- Add integration tests for K8s, Knative and Openshift
- Update documentation with new configuration

PR #875, opened 2022-02-24
Support Startup Probes for Kubernetes, Openshift and Knative

Sgitario:
The kubelet uses startup probes to know when a container application has started. If such a probe is configured, it disables liveness and readiness checks until it succeeds, making sure those probes don't interfere with the application startup. This can be used to adopt liveness checks on slow starting containers, avoiding them getting killed by the kubelet before they are up and running.

```yaml
startupProbe:
  httpGet:
    path: /healthz
    port: liveness-port
  failureThreshold: 30
  periodSeconds: 10
```

More context in [kubernetes.io/startup-probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).

Fix https://github.com/dekorateio/dekorate/issues/854


aureamunoz (approved):
LGTM

iocanel (approved):
LGTM

Issue #570, opened 2020-06-29
Add deploy support for knative


iocanel:
At the moment `-Ddekorate.deploy=true` doesn't seem to work properly.

PR #877, opened 2022-02-25
Add deploy and testing support for knative

Sgitario:
Updated documentation with the new JUnit extension for Knative.

Fix https://github.com/dekorateio/dekorate/issues/570


starters/knative/knative-junit-starter/pom.xml
```diff
@@ -14,7 +14,7 @@
   <dependencies>
```
Sgitario, line 14: Should we keep maintaining this `knative-junit-starter`? I don't think it makes sense any longer as users have know the `knative-junit` extension.

PR #879, opened 2022-02-28
Move Probe tests from `examples` to `tests`

Sgitario:
In this PR: https://github.com/dekorateio/dekorate/pull/875, we added the Probe coverage under the examples, where it should go under the tests modules.


iocanel (approved):
LGTM

Issue #857, opened 2022-02-04
Support gRPC probes for Kubernetes, Openshift and Knative


Sgitario:
# Description

Apart from the existing probes protocols HTTP, TCP and unix commands, in Kubernetes v1.23 the gRPC protocol can also be used. More context in [kubernetes.io/define-a-grpc-liveness-probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe).

If your application implements [gRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), kubelet can be configured to use it for application liveness checks. You must enable the GRPCContainerProbe [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) in order to configure checks that rely on gRPC.

Here is an example manifest:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: etcd-with-grpc
spec:
  containers:
  - name: etcd
    image: k8s.gcr.io/etcd:3.5.1-0
    command: [ "/usr/local/bin/etcd", "--data-dir",  "/var/lib/etcd", "--listen-client-urls", "http://0.0.0.0:2379", "--advertise-client-urls", "http://127.0.0.1:2379", "--log-level", "debug"]
    ports:
    - containerPort: 2379
    livenessProbe:
      grpc:
        port: 2379
      initialDelaySeconds: 10
```

# User story

```
Given an user who configures the new `dekorate.kubernetes.liveness-probe.grpcPort` property
When dekorate generates the kubernetes or openshift or Knative or other relevant manifests
Then the liveness probes should be generated using the new `grpc` section
```

```
Given an user who does NOT configure the new `dekorate.kubernetes.liveness-probe.grpcPort` property
When dekorate generates the kubernetes or openshift or other relevant manifests
Then the liveness probes should NOT be generated using the new `grpc` section
```

The new field `grpcPort` should be added in the `Probe` java class which is also used for `readiness`, `startup` (when implemented in https://github.com/dekorateio/dekorate/issues/854) probes.

# Acceptance criteria
- Add integration tests for K8s, Knative and Openshift
- Update documentation with new configuration

Sgitario:
Update: it seems that the gRPC probe will be available also in readiness, startup and liveness, not only in liveness: https://www.infoq.com/news/2021/12/kubernetes-1-23/

Description updated accordingly.

[...]

PR #880, opened 2022-02-28
Update Kubernetes Client dependency to most recent version 5.12.1

Sgitario:
All changes: https://github.com/fabric8io/kubernetes-client/compare/v5.10.0...v5.12.1

We need to bump the Kubernetes Client to the most recent version to support gRPC actions (see https://github.com/dekorateio/dekorate/issues/857)


aureamunoz (approved):

iocanel (approved):
LGTM

Issue #856, opened 2022-02-04
Support Job for Openshift and Kubernetes


Sgitario:
# Description

A Job creates one or more Pods and will continue to retry execution of the Pods until a specified number of them successfully terminate. As pods successfully complete, the Job tracks the successful completions. When a specified number of successful completions is reached, the task (ie, Job) is complete. Deleting a Job will clean up the Pods it created. Suspending a Job will delete its active Pods until the Job is resumed again.

More context in [kubernetes.io/jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/).

# User story

- If user does not set job configuration, then Job resource is NOT generated
- If user sets job configuration via properties or annotation, then Job resource is generated. 

## Properties

The Job spec is defined in [job-v1/#JobSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/#JobSpec).

| Property | Description | Mandatory |
| --- | --- | --- |
| name | name of the job | YES |
| restartPolicy | Restart policy when the job container fails. | NO. Default: OnFailure |
| parallelism | Specifies the maximum desired number of pods the job should run at any given time. | NO |
| completions | Specifies the desired number of successfully finished pods the job should be run with. | NO |
| completionMode | CompletionMode specifies how Pod completions are tracked. | NO. Default: NonIndexed |
| backoffLimit | Specifies the number of retries before marking this job failed. | NO. Default: 6 |
| activeDeadlineSeconds | Specifies the duration in seconds relative to the startTime that the job may be continuously active before the system tries to terminate it; value must be positive integer. | NO |
| ttlSecondsAfterFinished | ttlSecondsAfterFinished limits the lifetime of a Job that has finished execution (either Complete or Failed). If this field is set, ttlSecondsAfterFinished after the Job finishes, it is eligible to be automatically deleted.  | NO |
| suspend | Suspend specifies whether the Job controller should create Pods or not.  | NO. Default: true |
| containers[N].image | image of the container at position `N`  | YES |
| containers[N].name | name of the container at position `N`  | NO: If not set, Dekorate will use the image name |
| containers[N].command | image of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |
| containers[N].args | image of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |

## How do users configure the job configuration?

Let's see how would look like the job configuration for the mandatory fields only:

- Via properties
```
[...]

PR #881, opened 2022-02-28
Support Job for Openshift, Knative and Kubernetes

Sgitario:
More information in the feature request: https://github.com/dekorateio/dekorate/issues/856
Fix https://github.com/dekorateio/dekorate/issues/856


docs/documentation/kubernetes.md
```diff
@@ -292,6 +292,34 @@ Currently, the supported annotations for specifying volumes are:
...
+  }
+}
+```    
+
+Or via configuration properties at the file `application.properties`:
+
+    dekorate.kubernetes.jobs[0].name=say-hello
+    dekorate.kubernetes.jobs[0].containers[0].image=docker.io/user/hello
+
+Currently, the supported annotations for adding Kubernetes Jobs are:
```
aureamunoz, line 317:
```suggestion
Currently, the supported annotations for adding jobs are:
```

docs/documentation/kubernetes.md
```diff
@@ -292,6 +292,34 @@ Currently, the supported annotations for specifying volumes are:
...
+
+- @KubernetesApplication
+- @OpenShiftApplication
+- @KnativeApplication
```
aureamunoz, line 321: It should be nice to add this paragraph also to the Knative and OpenShift `docs/documentation/md` files, wdyt?

aureamunoz (changes requested):

iocanel (changes requested):
I am not sure if we should get into the business of defining arbitrary jobs via dekorate.
When I first heard about Job support in Kubernetes, I thought that the goal is to expose the application in development as a Job, rather than being able to define side jobs.

Let's meet and discuss.

Sgitario:
> I am not sure if we should get into the business of defining arbitrary jobs via dekorate. When I first heard about Job support in Kubernetes, I thought that the goal is to expose the application in development as a Job, rather than being able to define side jobs.
> 
> Let's meet and discuss.

As discussed, for Jobs, we're going to discuss two use cases:
- The generation of side Jobs. This is implemented as part of this PR. 
- Allow setting the kind output which will be implemented in https://github.com/dekorateio/dekorate/issues/886.

Sgitario:
@iocanel I've already updated this PR with the support of volumes.

iocanel (approved):
LGTM

PR #882, opened 2022-02-28
Support gRPC probes for Kubernetes, Openshift and Knative

Sgitario:
See feature request description in https://github.com/dekorateio/dekorate/issues/857
Fix https://github.com/dekorateio/dekorate/issues/857

### DIFF
--- a/assets/config.md
+++ b/assets/config.md
@@ -14,7 +14,7 @@ Each option can be:
 They can be referenced by index inside brackets as in java for example: `dekorate.kubernetes.labels[0]`.
 
 ## Complex object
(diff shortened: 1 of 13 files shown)


aureamunoz (approved):
LGTM! 
Shouldn't we add some doc about it? No sure if there is something about Probes but just wondering...

Sgitario:
> LGTM! Shouldn't we add some doc about it? No sure if there is something about Probes but just wondering...

Indeed! I missed updating the documentation. I've already done it. Thanks for spotting!

iocanel (approved):
LGTM

Issue #801, opened 2021-11-04
Allow to customise the values for `@OpenshiftIntegrationTest` via system properties


Sgitario:
When we launch a test using:

```java
@OpenshiftIntegrationTest
public class ManagedOpenShiftIT {
   @Inject
    Pod pod;
   // ...
}
```

The extension will build and deploy the application using the dekorate generated resources.

However, users might want to run the test in an `unmanaged` way, this is:

```java
@OpenshiftIntegrationTest(deployEnabled = false, buildEnabled = false)
public class UnmanagedOpenShiftIT {
   @Inject
    Pod pod;
  // ...
}
```

And the extension will keep resolving the instances like Pod, Utils, etc. 

Therefore, I think it would be useful to simply detect system properties (for example: `Ddekorate.test.deploy.enabled=false`) and use it. 
Moreover, I think it would be also helpful to have another flag to not delete the resources after running the test.

PR #884, opened 2022-03-02
test: allow to override testing configuration using system properties

Sgitario:
When we launch a test using:

```java
@OpenshiftIntegrationTest
public class ManagedOpenShiftIT {
   @Inject
    Pod pod;
   // ...
}
```

By default, the extension will build and deploy the application using the dekorate generated resources.

However, users might want to run the test in an `unmanaged` way, this is:

```java
@OpenshiftIntegrationTest(deployEnabled = false, buildEnabled = false)
public class UnmanagedOpenShiftIT {
   @Inject
    Pod pod;
  // ...
}
```

And the extension will keep resolving the instances like Pod, Utils, etc. 

With these changes, we can override the default test configuration using system properties, for example, for Kubernetes:

| Field                 | Type     | Description                                                                         | Overridable using System Property | Default Value |
|-----------------------|----------|-------------------------------------------------------------------------------------|-----------------------------------|---------------|
| deployEnabled         | boolean  | Flag to define whether the extension should automatically apply resources.          | dekorate.test.deploy.enabled      |          true |
| buildEnabled          | boolean  | Flag to define whether the extension should automatically perform container builds. | dekorate.test.build.enabled       |          true |
| readinessTimeout      | long     | The amount of time in milliseconds to wait for application to become ready.         | dekorate.test.readiness.timeout   |        300000 |
| additionalModules     | String[] | List of additional modules to be loaded by the test framework.                      | dekorate.test.additional-modules  |               |

Fix https://github.com/dekorateio/dekorate/issues/801

### DIFF
--- a/docs/documentation/junit-extensions.md
+++ b/docs/documentation/junit-extensions.md
@@ -91,7 +91,29 @@ This is important since in the `test` phase the application is not packaged. Her
 </plugin>
 ```
 
-#### related examples
+##### Integration Test Configuration
+
+We can customize the deployment configuration within the test execution using the Kubernetes integration test annotation:
+
+```java
+@KubernetesIntegrationTest(deployEnabled = true,
+  buildEnabled = true,
+  readinessTimeout = 300000,
+  additionalModules = {} 
+)
+class KubernetesIT {
+    // ...
+}
+```
+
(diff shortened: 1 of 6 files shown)


iocanel (approved):
LGTM

aureamunoz (approved):

Issue #886, opened 2022-03-03
Allow to change the default deployment type to either Job or CronJob or StatefulSet


Sgitario:
At the moment, when deploying a Kubernetes application, a `Deployment` resource will be automatically generated by Dekorate. However, we should allow users to define the concrete resource to generate between a range of resources like `StatefulSet` or `Job` resources.

This is already supported by Quarkus using the property `quarkus.kubernetes.deployment-kind`. See commit: https://github.com/quarkusio/quarkus/commit/4eaf76cee8c8deba03a8c4e719dd924b2b23c203

Issue #855, opened 2022-02-04
Support Kubernetes CronJob for Openshift and Kubernetes


Sgitario:
# Description

One [Cron](https://en.wikipedia.org/wiki/Cron)Job object is like one line of a crontab (cron table) file. It runs a job periodically on a given schedule, written in Cron format.

More context in [kubernetes.io/cron-jobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/).

# User story

- If user does not set cronjob configuration, then CronJob resource is NOT generated
- If user sets cronjob configuration via properties or annotation, then CronJob resource is generated. 

## Properties

The CronJob spec is defined in [cron-job-v1/#CronJobSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/cron-job-v1/#CronJobSpec).

|| Property || Description || Mandatory ||
| name | name of the cronjob | YES |
| schedule | cron expression | YES |
| concurrencyPolicy | cronjob concurrent policy | NO. Default: ALLOW |
| startingDeadlineSeconds | Optional deadline in seconds for starting the job if it misses scheduled time for any reason | NO |
| suspend | This flag tells the controller to suspend subsequent executions, it does not apply to already started executions. | NO. Default: false |
| successfulJobsHistoryLimit | The number of successful finished jobs to retain. | NO. Default: 3 |
| failedJobsHistoryLimit | The number of failed finished jobs to retain. | NO. Default: 1 |
| containers[N].name | name of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |
| containers[N].image | image of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |
| containers[N].command | image of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |
| containers[N].args | image of the container at position `N`  | NO: If not set, Dekorate could add the same one as in the Pod template? |

## How do users configure the cronjob configuration?

Let's see how would look like the cronjob configuration for the mandatory fields only:

- Via properties
```
dekorate.kubernetes.cronjobs[0].name=trigger-batch-process
dekorate.kubernetes.cronjobs[0].schedule=*/1 * * * *
```

- Via annotation
```java
@KubernetesApplication(cronJobs = {@CronJob(name = "trigger-batch-process", schedule = "*/1 * * * *")})
```

In this scenario, Dekorate would generate the following resource:

```yaml
apiVersion: batch/v1
[...]

PR #887, opened 2022-03-07
feat: allow to change the default deployment resource

Sgitario:
Fix https://github.com/dekorateio/dekorate/issues/886

Note, the full support of Job and CronJob will be done in https://github.com/dekorateio/dekorate/pull/881 and https://github.com/dekorateio/dekorate/issues/855.


iocanel (changes requested):
I would like the part that is related with the `getImageConfiguration` refactoring to be separated, as it seems to me more like a side improvement and not something that is actually needed by this PR. If I missed the actual connection, please elaborate.

Sgitario:
> getImageConfiguration

It's needed because we are calling it from the resource factory implementations, for example: https://github.com/dekorateio/dekorate/pull/887/files#diff-420e58003bb471a8fff038db7e321896fa49ee0434206162eddc96dc5db464e1R70. 

Note that resource factory can be called from any generator that inherits AbstractKubernetesManifestGenerator

iocanel (approved):
LGTM

PR #923, opened 2022-04-01
Fix typo in camel case example

glefloch:
In the `Casing` section of the docs, there is a typo in the camel case example.


iocanel (approved):
Great catch!

PR #926, opened 2022-04-04
feat: add java 14 and 17 in the build git action CI

aureamunoz:


.github/workflows/build.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [8,11]
+        java: [8,11,14,17]
```
iocanel, line 32: I would prefer to keep it `[8,11,17]` as it's easier to use those for local development with sdkman.

PR #936, opened 2022-04-12
feat: include helm model in helm annotations in noapt classifier

Sgitario:
Adding the model sources in the noapt classifier is necessary to overwrite the provided configuration.


iocanel (approved):
LGTM

PR #942, opened 2022-04-18
feat (#941): Common existing resources

iocanel:


Sgitario:
Is not this what the SessionReader is meant to do?

iocanel:
> Is not this what the SessionReader is meant to do?
You are right. There are files missing from the pull request. The question is how is it possible for the test to actually pass.

iocanel:
To answer my previous question ... The `tests/pom.xml` is also missing.

Sgitario (approved):

PR #950, opened 2022-04-26
refactor: upgrade supported java version to the latest 17

aureamunoz:


iocanel:
I am wondering if there is an alternative that would allow us to specify `any` release and avoid having these issues.

iocanel:
Unfortunately we can't set RELEASE_17 and expect it to work. What I propose instead is to follow this approach: https://github.com/kohsuke/metainf-services/pull/4/files.

I am closing this one and will create a new ASAP.

PR #951, opened 2022-04-27
fix: helm - update pom configuration for noapt and processors

Sgitario:
Fix this issue: When using the helm extension with the `noapt` classifier, it's failing because the listener is missing.


iocanel (approved):
LGTM

PR #959, opened 2022-04-28
fix: knative expose property

Sgitario:
According to https://knative.dev/docs/serving/services/private-services/, the right property is `networking.knative.dev/visibility`, not `serving.knative.dev/visibility`.


iocanel (approved):
LGTM

iocanel (approved):
LGTM

Issue #963, opened 2022-04-29
Container image name is not populated correctly using -Ddekorate.docker.group


cmoulliard:
## Issue

Version used: Dekorate 2.9-snapshot

The following command 
```
git fetch origin pull/949/head:pr-949
git checkout pr-949
mvn clean install -DskipTests
cd examples/spring-boot-with-certmanager-example

mvn clean install -Ddekorate.push=true \
   -Ddekorate.docker.registry=ghcr.io \
   -Ddekorate.docker.group=halkyonio \
   -Ddekorate.docker.name=halkyonio
```

still generates a wrong image name including my user's name and not `halkyonio`

```
Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
[INFO] Performing docker push.
The push refers to repository [ghcr.io/cmoullia/spring-boot-with-certmanager-example]
cdd2051c8a61: Preparing
55812b24d1c3: Preparing
138726990cf7: Preparing
5b40c5fb5220: Preparing
717b092b8c86: Preparing
denied: not_found: owner not found
```

That fails too using
```
mvn clean install \
    -Ddekorate.push=true \
    -Ddekorate.docker.image=ghcr.io/halkyonio/spring-boot-with-certmanager:1.0
The push refers to repository [docker.io/cmoullia/spring-boot-with-certmanager-example]
03fc6432284c: Preparing
55812b24d1c3: Preparing
138726990cf7: Preparing
5b40c5fb5220: Preparing
717b092b8c86: Preparing
denied: requested access to the resource is denied

```

PR #964, opened 2022-05-02
fix: populate Dekorate docker group and version from system properties

Sgitario:
Additionally, I've changed the annotation `@EnableDockerBuild` to `@DockerBuild` that was renamed in Dekorate 2.x.

Fix https://github.com/dekorateio/dekorate/issues/963

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,3 +51,12 @@ jobs:
           distribution: 'temurin'
       - name: Build Project
         run: ./mvnw ${MAVEN_ARGS} clean install
+      - name: Zip Artifacts
+        if: failure()
+        run: |
+          zip -R manifests-build-jvm${{ matrix.java }}.zip 'classes/META-INF/dekorate/*'
+      - uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: ci-manifests
+          path: manifests-build-jvm${{ matrix.java }}.zip
--- a/.github/workflows/integration-tests-with-dockerio.yml
+++ b/.github/workflows/integration-tests-with-dockerio.yml
@@ -87,6 +87,15 @@ jobs:
       - name: Install and Run Integration Tests
         run: |
           ./mvnw -B clean install -Presources -Pwith-examples -Pwith-tests -Pwith-service-binding-examples -Pwith-knative-examples -Duser.name=dekorateio -Dformat.skip=true -Ddocker-registry=docker.io
+      - name: Zip Artifacts
+        if: failure()
+        run: |
+          zip -R manifests-k8s-jvm${{ matrix.java }}.zip 'classes/META-INF/dekorate/*'
+      - uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: ci-manifests
+          path: manifests-k8s-jvm${{ matrix.java }}.zip
 
   openshift:
     name: Openshift Build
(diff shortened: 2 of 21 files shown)


cmoulliard (commented):
I did a test where I passed the version parameter and the result was wrong as the version tagged is `2.9-SNAPSHOT`
```
mvn clean install -Ddekorate.build=true -Ddekorate.docker.group=halkyonio -Ddekorate.docker.registry=ghcr.io -Ddekorate.docker.version=1.0
...
[INFO] Built image to Docker daemon as ghcr.io/halkyonio/spring-boot-on-kubernetes-with-jib-example:2.9-SNAPSHOT
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  6.941 s
[INFO] Finished at: 2022-05-03T10:46:10+02:00
[INFO] ------------------------------------------------------------------------
```

We have also the same issue using `mvn clean install -Ddekorate.build=true -Ddekorate.docker.image=ghcr.io/halkyonio/spring-boot-example:1.0`
-->
```
[INFO] Loading to Docker daemon...
[INFO] 
[INFO] Built image to Docker daemon as cmoullia/spring-boot-on-kubernetes-with-jib-example:2.9-SNAPSHOT
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

cmoulliard:
I did a test top of the commit `ebe8351c` and here is the result
```
OK  - mvn clean install -Ddekorate.build=true -Ddekorate.docker.group=halkyonio -Ddekorate.docker.registry=ghcr.io
OK  - mvn clean install -Ddekorate.build=true -Ddekorate.docker.group=halkyonio -Ddekorate.docker.registry=ghcr.io -Ddekorate.docker.version=1.0
NOK - mvn clean install -Ddekorate.build=true -Ddekorate.docker.image=ghcr.io/halkyonio/spring-boot-example:1.0
```

Sgitario:
> I did a test top of the commit `ebe8351c` and here is the result
> 
> ```
> OK  - mvn clean install -Ddekorate.build=true -Ddekorate.docker.group=halkyonio -Ddekorate.docker.registry=ghcr.io
> OK  - mvn clean install -Ddekorate.build=true -Ddekorate.docker.group=halkyonio -Ddekorate.docker.registry=ghcr.io -Ddekorate.docker.version=1.0
> NOK - mvn clean install -Ddekorate.build=true -Ddekorate.docker.image=ghcr.io/halkyonio/spring-boot-example:1.0
> ```
> 
> **NOTE**: If `dekorate.docker.image` is not longer supported, then this documentation page should be reviewed: https://dekorate.io/configuration-guide/#docker and also include an example part of the Description to explain what a group, registry are with same example.

This issue is only about allowing docker group and version properties to be set via system properties. These properties and also the docker registry property explicitly says in the docker documentation that can be set via system properties. 
In Dekorate, the rest of the properties (including docker image), unless is not mentioned in the documentation, cannot be set via system properties. So I don't think we should mention anything special for it. 

What could be good is to add a new column "can be set via system property or similar" in all the configuration tables to let users when we can set a property via system property. In any case, if we want to do this, this should be done in a separated issue.

cmoulliard:
> What could be good is to add a new column "can be set via system property or similar" in all the configuration tables to let users when we can set a property via system property

Good idea. I still dont understand by the way why we can set some properties as properties file or annotations and not  as `-Ddekorate.x.y.`. Can you explain ? @Sgitario

Sgitario:
> > What could be good is to add a new column "can be set via system property or similar" in all the configuration tables to let users when we can set a property via system property
> 
> Good idea. I still dont understand by the way why we can set some properties as properties file or annotations and not as `-Ddekorate.x.y.`. Can you explain ? @Sgitario

I've already created https://github.com/dekorateio/dekorate/issues/965 to do this.

iocanel (approved):
LGTM

cmoulliard (approved):

PR #966, opened 2022-05-06
fix: run the checks job on push, instead of in pull requests

Sgitario:
This is because some links are not available until they are available in the main branch.

Related failure that we'll about with this change: https://github.com/dekorateio/dekorate/runs/6290448335?check_suite_focus=true


iocanel (approved):
LGTM

PR #971, opened 2022-05-12
docs: fix anchors in the index page

Sgitario:
For example, the `Dekorate 2.9.3` text was displaying an anchor what was not expected to have.
I've added a CSS class with name `no-anchors`, so we can use it to skip the anchors in other places.


iocanel (approved):
LGTM

Issue #965, opened 2022-05-06
Populate Dekorate configuration from user properties and system properties


Sgitario:
# Summary

Allow to provide Dekorate configuration from:

- Annotations. Example:
```java
@KubernetesApplication(replicas = 3)
class MyApplication {
// ...
}
```

- Properties. Example:
```
dekorate.kubernetes.replicas=3
```

- And now system properties as well. For example, when we build our application using Maven:
```
mvn clean install -Ddekorate.kubernetes.replicas=3
```

# Implementation
Change the method `propertiesMap` in https://github.com/dekorateio/dekorate/blob/646f9dacc8428e4a9fb0f31865e41f1da0e0d93f/core/src/main/java/io/dekorate/SessionHandler.java#L49 to also ingest the dekorate configuration from the system properties.

# More context
This issue is a follow-up action to this comment: https://github.com/dekorateio/dekorate/pull/964#issuecomment-1116993770

cmoulliard:
Can you please describe what problem we will fix if we do this change ? @Sgitario

Sgitario:
> Can you please describe what problem we will fix if we do this change ? @Sgitario

Updated description.

PR #972, opened 2022-05-12
Populate Dekorate configuration from user properties and system properties

Sgitario:
These changes are already tested and covered by `tests/issue-963-docker-system-properties` (see changes).

Fix https://github.com/dekorateio/dekorate/issues/965


core/src/main/java/io/dekorate/Session.java
```diff
@@ -166,19 +166,19 @@ public void addConfiguration(Map<String, Object> map, BiConsumer<ConfigurationGe
...
-      if (value instanceof Map) {
-        Map<String, Object> generatorMap = new HashMap<>();
-        Class configClass = configtypes.get(key);
-        String newKey = configClass.getName();
-        Generators.applyPrimitives(configClass, (Map<String, Object>) value);
-        Generators.populateArrays(configClass, (Map<String, Object>) value);
-        generatorMap.put(newKey, value);
-        consumer.accept(generator, Maps.kebabToCamelCase(generatorMap));
+      if (generator != null) {
+        if (value instanceof Map) {
```
Sgitario, line 170: I had to make this change because, during the build, we're adding some `dekorate.docker.xxx` system properties. As part of this PR, we're now including the system properties, therefore some examples were failing because the `docker` generator didn't exist (the `docker-annotations` module was not in the Maven configuration).

Having said that, I do think the right behavior is to not fail, but simply add a warning in the logs to let users know. 
What do you think about this change? @iocanel

iocanel (commented):
Overall, I like it.

We do need to filter incoming System properties.
And I am a bit skeptic about relaxing the error about missing generators.

core/src/main/java/io/dekorate/processor/AbstractAnnotationProcessor.java
```diff
@@ -116,6 +131,10 @@ protected Map<String, Object> readApplicationConfig(String... resourceNames) {
     return result;
   }
 
+  private Map<String, Object> readDekorateSystemProperties() {
+    return Maps.fromProperties(System.getProperties());
```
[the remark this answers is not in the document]
iocanel: Didn't remember that!

core/src/main/java/io/dekorate/Session.java
iocanel: I think that this brings to the surface that problems that arise from not having profile support or something like this.
So, I'd say it's fair to relax this check.

iocanel (approved):
LGTM

PR #977, opened 2022-05-13
fix: delete logging configuration in helm annotation dependency

Sgitario:
With this logging configuration, the spring boot applications traces were turned off.


Sgitario:
I will delete the JSONPath expressions support (will provide something similar), and also the JAYWAY dependency, so this change is not necessary.

PR #978, opened 2022-05-13
fix: work around AnnotationTypeMismatchException.

iocanel:
This pull request addresses the issue with the `AnnotationTypeMismatchException`. that occurs in some cases using JDK 11+ when trying to read the Ingress from the KubernetesApplication.

The workaround catch the `Exception` (`AnnotationTypeMismatchException` seems uncatchable, maybe it's proxied or something like this) and returns the default value for the method. As the issue seems to only take place when no explicit value is specified falling back to the default return value seems good enough.

### DIFF
--- a/core/src/main/java/io/dekorate/utils/Maps.java
+++ b/core/src/main/java/io/dekorate/utils/Maps.java
@@ -156,20 +156,23 @@ public class Maps {
     return result;
   }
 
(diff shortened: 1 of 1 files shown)


Sgitario (approved):
I've verified that this fix also solves the issue that I was having about annotation type mismatch exceptions.

PR #980, opened 2022-05-16
feat: align ingress configuration between Dekorate and Quarkus

Sgitario:
In Quarkus, the ingress configuration is set as:
```
quarkus.kubernetes.ingress.expose=true
```

However, in Dekorate, it is:
```
dekorate.kubernetes.expose=true
```

This pull request aligns both configurations and uses the one from Quarkus.


Sgitario:
Next actions (in separated pull requests):
- Do the same for OpenShift (routes)
- Add the annotations field in the Ingress and Route resources (as in Quarkus)
- Do the same for Knative
- Allow using Ingress resources in OpenShift

iocanel (changes requested):
Overall looks good.Can we apply that to Openshfit too? And maybe rename the annotation from Ingress to Expose so that its not to Vanilla specific?

Sgitario:
I planned to do the OpenShift route thing as part of a separate pull request. See the next actions https://github.com/dekorateio/dekorate/pull/980#issuecomment-1127235869

About renaming Ingress to simply Expose, I disagree with this. Expose is a too abstract concept and it would go in the direction of being unaligned again to what we have in Quarkus. Moreover, we lose the opportunity to add Ingress-specific attributes like `tls` in here.

iocanel (approved):
LGTM

iocanel:
@Sgitario: This is approved but there is a conflict to resolve before we merge.

Sgitario:
> @Sgitario: This is approved but there is a conflict to resolve before we merge.

Conflicts resolved. Waiting for CI to be green.

PR #981, opened 2022-05-16
fix: Replace jayway JsonPath by custom YAML expressions

Sgitario:
Due to Jayway JsonPath dependency introduces some unnecessary dependencies at runtime and also print some annoying traces, we have decided to get rid of the support of JsonPath. 

To replace JsonPath, we've introduced a custom YAML expressions very similar to JsonPath, but without needing to parse from/to YAML to Json. 

Documentation has been updated

### DIFF
--- a/annotations/helm-annotations/pom.xml
+++ b/annotations/helm-annotations/pom.xml
@@ -18,11 +18,6 @@
     </dependency>
 
     <dependency>
-      <groupId>com.jayway.jsonpath</groupId>
-      <artifactId>json-path</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>io.sundr</groupId>
       <artifactId>builder-annotations</artifactId>
       <scope>compile</scope>
--- a/annotations/helm-annotations/src/main/java/io/dekorate/helm/annotation/ValueReference.java
+++ b/annotations/helm-annotations/src/main/java/io/dekorate/helm/annotation/ValueReference.java
@@ -28,11 +28,9 @@ public @interface ValueReference {
   String property();
 
   /**
-   * The JSONPath expressions where to map the property.
-   *
-   * It uses <a href="https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html">JSONPath expression specification</a>.
+   * The path expressions where to map the property.
    */
-  String[] jsonPaths();
+  String[] paths();
 
   String profile() default "";
 
--- a/annotations/helm-annotations/src/main/java/io/dekorate/helm/listener/HelmWriterSessionListener.java
+++ b/annotations/helm-annotations/src/main/java/io/dekorate/helm/listener/HelmWriterSessionListener.java
@@ -14,7 +14,6 @@
  * limitations under the License.
  *
  **/
-
 package io.dekorate.helm.listener;
 
 import static io.dekorate.helm.config.HelmBuildConfigGenerator.HELM;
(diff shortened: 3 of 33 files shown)


iocanel (changes requested):
My main concern is the fact that the pull request is drastically changing the user experience, in order to workaround technical issues. So, it feels like a convinience change.

+1: for removing jayway (I agree that we should not use external dependencies)
-1: for introducing a custom expression language.

The main reason I prefer `json-path` is familiarity. It's easier to describe and explain to people.
And yes, in many ways it's an overkill and might be a little too verbose, but from my perspective it lifts from our shoulders the burden of teaching users a new language. A propriatary language starts bringing up questions...

For example:
- Do we support arrays? 
- Do we support filtering by name or index?

Now, I don't want to enter an endless debate, so if you feel that this is the way to go, then please provide addtional documentation and tests so that we at least cover cases like named containers, volumes, mounts etc.

Sgitario:
> My main concern is the fact that the pull request is drastically changing the user experience, in order to workaround technical issues. So, it feels like a convinience change.
> 
> +1: for removing jayway (I agree that we should not use external dependencies) -1: for introducing a custom expression language.
> 
> The main reason I prefer `json-path` is familiarity. It's easier to describe and explain to people. And yes, in many ways it's an overkill and might be a little too verbose, but from my perspective it lifts from our shoulders the burden of teaching users a new language. A propriatary language starts bringing up questions...
> 
> For example:
> 
> * Do we support arrays?
> * Do we support filtering by name or index?
> 
> Now, I don't want to enter an endless debate, so if you feel that this is the way to go, then please provide addtional documentation and tests so that we at least cover cases like named containers, volumes, mounts etc.

Yes, it supports arrays and filters by name. 
I do think this is the way to go, plus I've already updated the existing documentation about its usage and also migrated all the examples (that uses the expressions in many different forms). 

I've re-requested your review.

iocanel (changes requested):
Figuring out how the expression language works is not trivial, so we need to be very specific about what we support and what not. If we are to introduce a new expression language, we need to teach users how to use it. So, this part of the doc should times more detailed than anything else we 've documented so far.

I still don't see in the doc, how to refer to container by name. The doc should contain such examples. Some goes, for more complex cases like container ports by name.
Container environment varialbes is also something I'd like to see.

I also want us to describe if the the expresssion support any kind of wildchars. If we don't we need to let users know.

Sgitario:
> Figuring out how the expression language works is not trivial, so we need to be very specific about what we support and what not. If we are to introduce a new expression language, we need to teach users how to use it. So, this part of the doc should times more detailed than anything else we 've documented so far.
> 
> I still don't see in the doc, how to refer to container by name. The doc should contain such examples. Some goes, for more complex cases like container ports by name. Container environment varialbes is also something I'd like to see.
> 
> I also want us to describe if the expression support any kind of wildchars. If we don't we need to let users know.

I've added some more examples about how to map containers by name (I've added also some unit tests), and also more about the filters we can write (I've also implemented the OR conditions). The example about container environment variables is similar to the annotations example (it's a map). 

Moreover, I've included another section about what is not supported (we could potentially implement some unsupported features in the future if needed).

From my point of view, with these changes, the expression language is simple and documented enough as it is. So, users can easily learn how to use it. Having said that, we can always improve the path expressions and/or the documentation from users' feedback. 

If you don't mind, as this pull request fixes a critical bug when using the Dekorate helm dependency in Spring Boot applications, then when CI is green, I will merge it. We can always improve it in other pull requests.

Issue #976, opened 2022-05-13
Helm: Allow to define the root name in the `Values.yaml` file


Sgitario:
We are using the module name to map the values. For example, if our module name was "spring-boot-on-kubernetes-example", the generated `Values.yaml` file by the Dekorate helm extension had:

```
springBootOnKubernetesExample:
  replicas: 3
  ...
```

From my point of view, this is not very user-friendly... let's add a new property called `dekorate.helm.root-alias`, in which the default value is `app`. So, we won't use the module name any longer, but "app", plus it will allow users to provide a different value.

With these changes, the generate Values file would be:

```
app:
  replicas: 3
  ...
```

PR #986, opened 2022-05-19
feat: provide root name for aliases

Sgitario:
fix: https://github.com/dekorateio/dekorate/issues/976


annotations/helm-annotations/src/main/java/io/dekorate/helm/listener/HelmWriterSessionListener.java
```diff
@@ -85,6 +85,7 @@ public void onClosed() {
     Session session = getSession();
     Project project = getProject();
     Path outputDir = project.getBuildInfo().getClassOutputDir().resolve(project.getDekorateOutputDir());
+    System.out.println("Project name: " + project.getBuildInfo().getName());
```
iocanel, line 88: I think we need to remove this or switch to the Logger.

iocanel (approved):
LGTM

annotations/helm-annotations/src/main/java/io/dekorate/helm/listener/HelmWriterSessionListener.java
Sgitario: Ops, removed!

Issue #995, opened 2022-06-04
Invalid warnings appear in the log


iocanel:
Running `mvn clean package -Ddekorate.build=true` produces:
```
[WARN] Unknown generator 'build' will be ignored. Known generators are: [jvm, servicemonitor, vcs, options, knative, docker].
```

In the same manner: `mvn clean package -Ddekorate.push=true` produces:
```
[WARN] Unknown generator 'docker' will be ignored. Known generators are: [jvm, servicemonitor, vcs, options, knative, docker].
```

`build`, `push` and `deploy` should never produce warnings. We either need to add a check for them or properlly represent them using generators.

iocanel:
@Sgitario can you please have a look?

Sgitario:
I guess the second code snippet is wrong, the right one should be:

```
[WARN] Unknown generator 'push' will be ignored. Known generators are: [jvm, kubernetes, servicemonitor, vcs, options, docker]
```

Moreover, this affects all the properties that are not mapped using generators like `dekorate.input.dir`, `dekorate.output.dir`, `dekorate.image-pull-secrets`, `dekorate.properties.profile`, `dekorate.verbose`, `dekorate.test.xxx`, ... and maybe more coming in the future.

I will change the logic of this method to not print traces at all.

Sgitario:
Unassign this issue. See comment in https://github.com/dekorateio/dekorate/pull/997#issuecomment-1148300050. 
If I have the time in the future, I will think to take this one back to me. However, keep it unassigned just in case somebody else has the capacity to fix it.

PR #997, opened 2022-06-07
fix: invalid warnings when running the Dekorate session

Sgitario:
There are some system properties that are not directly mapped using generators like `dekorate.build` or `dekorate.push`. These properties will either enable some special handling, functionality and/or include some other properties. 

The problem is that at the moment, the session is checking whether these properties are properly mapped and if not, it will trace a warning. 

Therefore, as these properties are programmatically used and there are many more like `dekorate.input.dir`, `dekorate.output.dir`, `dekorate.image-pull-secrets`, `dekorate.properties.profile`, `dekorate.verbose`, `dekorate.test.xxx`, ... and maybe more coming in the future, I think the best way is to not trace this warning ever. 

The alternative is to maintain a list with all the above values, however I don't really like this approach as this list will be likely became out-date when adding a new unmapped property.

Fix https://github.com/dekorateio/dekorate/issues/995

### DIFF
--- a/core/src/main/java/io/dekorate/Session.java
+++ b/core/src/main/java/io/dekorate/Session.java
@@ -162,23 +162,17 @@ public class Session {
   }
 
   public void addConfiguration(Map<String, Object> map, BiConsumer<ConfigurationGenerator, Map<String, Object>> consumer) {
-    for (Map.Entry<String, Object> entry : map.entrySet()) {
-      String key = entry.getKey();
-      Object value = entry.getValue();
-      ConfigurationGenerator generator = configurationGenerators.get(key);
-      if (generator != null) {
-        if (value instanceof Map) {
-          Map<String, Object> generatorMap = new HashMap<>();
-          Class configClass = configtypes.get(key);
-          String newKey = configClass.getName();
-          Generators.applyPrimitives(configClass, (Map<String, Object>) value);
-          Generators.populateArrays(configClass, (Map<String, Object>) value);
-          generatorMap.put(newKey, value);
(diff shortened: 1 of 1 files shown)


iocanel:
Not providing feedback to the user because it's not convinient does not seem right. 

As described in #995: I would either handle those cases or even ensure that the properties are properlly backed by configuration objects.

BTW, why didn't we have this issue before: https://github.com/dekorateio/dekorate/commit/62aab5b4c0807cc95f7bcf3449c035076d98dcef

Sgitario:
> Not providing feedback to the user because it's not convinient does not seem right.
> 
> As described in #995: I would either handle those cases or even ensure that the properties are properlly backed by configuration objects.
> 
> BTW, why didn't we have this issue before: [62aab5b](https://github.com/dekorateio/dekorate/commit/62aab5b4c0807cc95f7bcf3449c035076d98dcef)

Before, it was raising exceptions when a configuration was not using a generator, not by a warning trace. 

I don't totally agree with you about this is something we should tell users. 
As you said maybe the best way to handle these unmapped properties is to actually map them using configuration objects, however as there are many unmapped properties, this is going to become a huge task. So, I will close this pull request and unassign the issue as I don't have this time right now.

iocanel:
> Before, it was raising exceptions when a configuration was not using a generator, not by a warning trace.

Indeed, still no error was thrown when `-Ddekorate.build=true` or `-Dekorate.push=true` was passed. 
Anyway, I prefer getting this right even if it is a huge task.

PR #999, opened 2022-06-07
chore: remove dekorate.input/output.dir in favor of dekorate.options.…

iocanel:
…input/output-path


annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/config/DefaultKubernetesConfigGenerator.java
```diff
@@ -29,7 +28,6 @@ public class DefaultKubernetesConfigGenerator implements KubernetesConfigGenerat
   public DefaultKubernetesConfigGenerator(ConfigurationRegistry configurationRegistry) {
     this.configurationRegistry = configurationRegistry;
     this.configurationRegistry.add(new ApplyProjectInfo(getProject()));
-    this.configurationRegistry.add(new ApplyImagePullSecretConfiguration());
```
aureamunoz, line 32: Why does this configurator deisappear?

iocanel: It is no longer needed. Ever since we added automatic mapping of system properties to config objects.

annotations/option-annotations/src/main/java/io/dekorate/option/apt/GeneratorOptionsProcessor.java
```diff
@@ -37,9 +37,6 @@
 })
 public class GeneratorOptionsProcessor extends AbstractAnnotationProcessor implements WithSession {
 
-  private static final String INPUT_DIR = "dekorate.input.dir";
-  private static final String OUTPUT_DIR = "dekorate.output.dir";
-
   private static final String FALLBACK_INPUT_DIR = "META-INF/fabric8";
   private static final String FALLBACK_OUTPUT_DIR = "META-INF/fabric8";
```
aureamunoz, line 41: It seems that those are not used, right? Can we remove them?

Issue #1006, opened 2022-06-29
Enum default values should not be forced.


iocanel:
Currently in dekorate we have almost 18 enums. Those enums are used in the dekorate annotations. In most cases the annotations define the default value and the rest of the code just applies whatever value is found in the config objects. This works fine for dekorate itself but has issues with downstream projects. Why? because a dekorator applying the default or overriden value will always be applied possibly clashing with dekorators registered downstream.

For example, with the given code we always register a dekorator that sets the image pull policy, wether the user selected a non default value or not. This is breaking downstream projects that use their own config model.

What should we do? Only register decorators when non default values are selected.

Note: This issue most probably applies to other types too.

PR #1008, opened 2022-06-29
fix: leverage builderWithDefaults to decouple handling of default values

iocanel:
This is related to #1006 

The pull request introduces static config objects with default values for BaseConfg, KubernetesConfig, KnativeConfig in order to maintain an instance of the default configuration so that we only apply dekorators when non-default values are used.

For now, the issue covers `ImagePullPolicy` and also global/revision autoscaling of knative which seem to be the most common cases of probelms.

### DIFF
--- a/core/src/main/java/io/dekorate/AbstractKubernetesManifestGenerator.java
+++ b/core/src/main/java/io/dekorate/AbstractKubernetesManifestGenerator.java
@@ -21,7 +21,6 @@ import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
-import io.dekorate.kubernetes.annotation.ImagePullPolicy;
 import io.dekorate.kubernetes.config.Annotation;
 import io.dekorate.kubernetes.config.AwsElasticBlockStoreVolume;
 import io.dekorate.kubernetes.config.AzureDiskVolume;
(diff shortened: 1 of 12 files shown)


iocanel:
This will also deal with issues reported on Quarkus by @spolti

spolti (approved):

Sgitario (changes requested):
If I understood well the issue, the problem is that the default value from annotations is ignored when mapping the annotations to the actual properties? If this is right, I would look into a more general solution that also fixes this for Kubernetes, OpenShift, and other generators.

Also, I think either an integration test or simply a unit test should be added to reproduce the issue and confirm it's solved.

iocanel:
> If I understood well the issue, the problem is that the default value from annotations is ignored when mapping the annotations to the actual properties?

Not exactly, Default values are respected fine. The issue is that we apply dekorators even when the default value is present in the config objects. Again, this is desired in dekorate land, but if dekorate is used downstream without dekorate's own config model, this is causing issues.

> Also, I think either an integration test or simply a unit test should be added to reproduce the issue and confirm it's solved.

An itnegration test is not possible, as in the dekorate land there is no problem. I'll check to see if we can have a unit test that checks if the registered dekorators are the expected ones.

iocanel:
@Sgitario: Added ingtegration test that verify that the dekorators are only registrered when non default options are provided. How? By providing existing resources with non-default configuration and checking that they are respected.

core/src/main/java/io/dekorate/AbstractKubernetesManifestGenerator.java
```diff
@@ -111,7 +113,7 @@ protected void addDecorators(String group, C config) {
       resourceRegistry.decorate(new ApplyServiceAccountNamedDecorator(config.getName(), config.getServiceAccount()));
     }
 
-    if (config.getImagePullPolicy() != ImagePullPolicy.IfNotPresent) {
+    if (config.getImagePullPolicy() != DEFAULT_BASE_CONFIG.getImagePullPolicy()) {
       resourceRegistry.decorate(group, new ApplyImagePullPolicyDecorator(config.getImagePullPolicy()));
```
Sgitario, line 117: missing the first parameter `config.getName()` here

Sgitario (commented):
Thanks for adding the test. I see what the issue is resolving now. 
There are some failures, apart from it, the changes look good to me.

Sgitario (approved):

PR #1011, opened 2022-07-11
Use `yaml-path` dependency in Helm annotations

Sgitario:
Link to the yaml path library: https://github.com/yaml-path/YamlPath

The tests are being deleted because they are now part of the library.


Sgitario:
Closing with won't do.

PR #1013, opened 2022-07-11
Use `yaml-path` dependency in Helm annotations

Sgitario:


Sgitario:
@iocanel what about merging this one for Dekorate 2.x?

PR #1014, opened 2022-07-12
Merge properties from system properties and application properties

Sgitario:
Before these changes, if user provides a system property, for example: `dekorate.options.push=true`, and having another property in the `application.properties` file, for example `dekorate.options.input-path=k8s`, Dekorate ignored the properties from the application properties and only takes the one from the system properties.


iocanel (approved):
LGTM

Issue #1017, opened 2022-07-13
Maven build uses "src/main/resources" to read the application properties instead of "target/classes"


Sgitario:
By default, Maven copies all the resources in "src/main/resources" to "target/classes". 
However, when using the Maven resources plugin, we can alter the application properties (and all the files) by filtering. For example, having the following properties:

```
app.value=${SOME_PROPERTY}
```

And using the Maven resource plugin as:

```xml
<build>
  <resources>
    <resource>
      <directory>src/main/resources</directory>
      <filtering>true</filtering>
    </resource>
  </resources>
</build>
```

Running the following Maven command: "mvn clean install -DSOME_PROPERTY=here", Maven will replace the property SOME_PROPERTY to:

```
app.value=here
```

And it will copy the updated application properties into the folder "target/classes".

Therefore, Dekorate should always use "target/classes" instead of "src/main/resources".

Issue #1018, opened 2022-07-13
The Dekorate options input-path is ignored when using the Maven resources plugin


Sgitario:
When using the Maven resources plugin, for example:

```xml
<build>
  <resources>
    <resource>
      <directory>any-folder</directory>
      <filtering>true</filtering>
    </resource>
  </resources>
</build>
```

And using the Dekorate options input-path:

```
dekorate.options.input-path=k8s
```

Even though the resources do exist, they are ignored.

I need to investigate a bit more why this is happening.

PR #1020, opened 2022-07-13
Change Maven resources dir from "src/main/resources" to "target/classes"

Sgitario:
Fix https://github.com/dekorateio/dekorate/issues/1017


Sgitario:
Added another commit to prove that the Maven resources filtering feature work fine when importing custom resources.

Closes https://github.com/dekorateio/dekorate/issues/1018

iocanel (approved):
LGTM

PR #1024, opened 2022-07-13
fix: knative was generating an empty ConfigMap

Sgitario:
This problem was spotted by https://github.com/dekorateio/dekorate/actions/runs/2663085841

Failing test: KnativeGlobalContainerConcurrencyTest

### DIFF
--- a/annotations/knative-annotations/src/main/java/io/dekorate/knative/manifest/KnativeManifestGenerator.java
+++ b/annotations/knative-annotations/src/main/java/io/dekorate/knative/manifest/KnativeManifestGenerator.java
@@ -87,6 +87,9 @@ import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 
 public class KnativeManifestGenerator extends AbstractKubernetesManifestGenerator<KnativeConfig> {
 
+  public static final String CONFIG_AUTOSCALER = "config-autoscaler";
+  public static final String CONFIG_DEFAULTS = "config-defaults";
+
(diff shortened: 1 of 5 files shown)


annotations/knative-annotations/src/main/java/io/dekorate/knative/manifest/KnativeManifestGenerator.java
```diff
@@ -211,25 +212,25 @@ public void generate(KnativeConfig config) {
 
     // Global autoscaling configuration
     if (config.getGlobalAutoScaling() != null) {
-      if (!isDefault(config.getGlobalAutoScaling())) {
+      if (config.getGlobalAutoScaling().getAutoScalerClass() != null
+          && config.getGlobalAutoScaling().getAutoScalerClass() != AutoScalerClass.kpa) {
         resourceRegistry.decorate(KNATIVE, new AddConfigMapResourceProvidingDecorator(CONFIG_AUTOSCALER, KNATIVE_SERVING));
```
Sgitario, line 217: Before these changes, we were creating the `config-autoscaler` configmap, when the property `target-utlization-percentage` is set which does not use it. 

I've removed the `isDefault` method and create the configmap if and only if it's necessary.

iocanel (changes requested):
I would like pull requests to be a little more focused.
- It's really hard for someone looking at the pull request to understand what the root cause was.
- It's a a tad more drastatic than it needs too, which could in theory cause other kind of regressions (BTW It doesn't, but I had to check to make sure).

iocanel (approved):
I was just seeing half of the issue.
So, the PR is not as `drastic` as it initally felt. Looks good.

PR #1031, opened 2022-08-08
Helm: Ensure deterministic order when applying config references

Sgitario:
This issue has been spotted in https://github.com/quarkiverse/quarkus-helm/issues/41 and after these changes, now it worked fine. 

It ensures deterministic order and will always use the last value found.


Sgitario:
Adding a test would be a bit tricky (however will cover this change in the Quarkiverse extension).

aureamunoz (commented):
Is there a test for this?

Sgitario:
> Is there a test for this?

No, as this is a random issue when using different decorators and the same properties.

aureamunoz (approved):
LGTM

PR #1032, opened 2022-08-11
Support emptydir volumes

Sgitario:
We support many volume types, but we're missing the emptydir volume. This volume is a request from the user community: https://github.com/quarkusio/quarkus/issues/27222


iocanel (approved):
LGTM

PR #1037, opened 2022-08-30
simplifying SB examples

aureamunoz:
I thing that keeping these examples as simple as possible only with dekorte dependency is a good idea, WDYT?


iocanel (approved):
LGTM

PR #1052, opened 2022-09-05
Helm: Add new properties `conditions` and `tags`

Sgitario:
These two properties are new in Helm chart API version v2: https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags

This is necessary for https://github.com/quarkiverse/quarkus-helm/issues/67


iocanel (approved):
LGMT

PR #1058, opened 2022-09-12
Use named port for OpenShift routes

Sgitario:
This is related to this issue: https://github.com/quarkusio/quarkus/issues/26587


Sgitario:
PR updated with:
- Create a `Route` config in OpenShift application
- Move `expose` and `host` properties from OpenshiftApplication to the new `Route`
- Create a new property `targetPort` property in Route

iocanel (approved):
lgtm

PR #1062, opened 2022-09-16
fix: provide the config references in helm before running the helm stage

Sgitario:
This is necessary to overcome the Dekorate limitation where we cannot use Object fields in annotations.

This way we can allow other extensions like Quarkus Helm supports this.


Sgitario:
Related to https://github.com/quarkiverse/quarkus-helm/issues/83#issuecomment-1249160328

PR #1072, opened 2022-10-05
Fix Kind extension to generate manifests

Sgitario:
Plus, add the Kind page into the dekorate site. 
This pull request is related to https://github.com/quarkusio/quarkus/issues/28078

### DIFF
--- a/docs/documentation/kind.md
+++ b/docs/documentation/kind.md
@@ -38,7 +38,7 @@ This module can be added to the project using:
 </dependency>
 ```
 
-The `@Kind` annotation can be used in combination with `@KubernetesApplication`. In that case the `@Kind` configuration will be replaced by the one specified using `@KubernetesApplication`.
+The `@Kind` annotation can be used in combination with `@KubernetesApplication`. In that case the `@Kind` configuration will take precedence over the one specified using `@KubernetesApplication` in the generated Kind manifests.
 
 **REMINDER**: A complete reference on all the supported properties can be found in the [configuration options guide]({{site.baseurl}}/configuration-guide).
 
--- a/docs/documentation/kind.md
+++ b/docs/documentation/kind.md
@@ -1,0 +1,45 @@
+---
+title: Kind
+description: Kind
+layout: docs
+permalink: /docs/kind
+---
+
+### Kind
+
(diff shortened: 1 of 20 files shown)


iocanel (changes requested):
The pull request looks good.
But I think that besides the `imagePullPolicy` we should also provide a default value to `serviceType`. More specifically we need to set it by default to `NodePort` for kind yml/json.

This is something that we already do for minikube/kind in Quarkus and it would be good to keep things aligned.

Sgitario:
PR updated including the suggested change.

iocanel (approved):
LGTM

examples/kind-example/pom.xml
```diff
@@ -0,0 +1,106 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>io.dekorate</groupId>
+  <artifactId>kind-example</artifactId>
+  <version>3.1-SNAPSHOT</version>
+  <name>Dekorate :: Examples :: Kind on Kubernetes</name>
```
aureamunoz, line 8: I would name it `Kind cluster` or something like this, instead of Kind on Kubernetes because Kind is already Kubernetes (in docker)

docs/documentation/kind.md
```diff
@@ -0,0 +1,45 @@
...
+
+```xml
+<dependency>
+  <groupId>io.dekorate</groupId>
+  <artifactId>kind-annotations</artifactId>
+  <version>{{site.data.project.release.current-version}}</version>
+</dependency>
+```
+
+The `@Kind` annotation can be used in combination with `@KubernetesApplication`. In that case the `@Kind` configuration will be replaced by the one specified using `@KubernetesApplication`.
```
aureamunoz, line 41: Shouldn't it be the opposite? Shouldn't @Kind take precedence if it's a more Kubernetes specialization?

examples/kind-example/pom.xml
Sgitario: very good point

docs/documentation/kind.md
Sgitario: Indeed. I will fix it in the minikube documentation (this is a copy paste :)

Sgitario:
@aureamunoz PR updated

PR #1082, opened 2022-10-17
docs: fix documentation about the property `properties-profile`

Sgitario:
The correct property is `dekorate.options.properties-profile`, not `dekorate.properties-profile`.  This changed recently when introducing the wrap config Options.  I've also fixed the example using this property to correctly cover this functionality.


iocanel (approved):
lgtm

PR #1092, opened 2022-10-25
Prevent generating duplicate Ingress paths rules

Sgitario:
Related to https://github.com/quarkusio/quarkus/issues/28812


annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/decorator/AddIngressRuleDecorator.java
```diff
@@ -15,12 +15,8 @@
  */
 package io.dekorate.kubernetes.decorator;
 
-import static io.dekorate.kubernetes.decorator.AddServiceResourceDecorator.distinct;
```
iocanel, line 18: I `think` that distinct should move to a utility class.

iocanel (approved):
LGTM
Added a comment about the use of `distinct` class. This is something that its good to address at some point, but not a blocker.

ThoSap:
Could it be that this did not make it into Quarkus `2.14.0.Final` but https://github.com/quarkusio/quarkus/pull/28813 did?

PR #1107, opened 2022-12-05
Fix integration tests failures in Kubernetes and OpenShift

Sgitario:
The Kubernetes tests are fixed by using manusa/actions-setup-minikube@v2.7.1
However, the OpenShift tests are still failing :/. Any help here @manusa ?

### DIFF
--- a/.github/workflows/integration-tests-with-dockerio.yml
+++ b/.github/workflows/integration-tests-with-dockerio.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2.3.4
       - name: Setup Minikube-Kubernetes
-        uses: manusa/actions-setup-minikube@v2.4.1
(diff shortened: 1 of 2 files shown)


manusa:
Hi, just saw this one.
I'm not sure where else is the actions-setup-openshift used.
GitHub updated the ubuntu-latest to be based on 22.04 since the 1st of december. Unfortunately the action won't work in this version of ubuntu. Need to change the runs-on declaration to `ubuntu-20.04` that should fix it.

Sgitario:
> Hi, just saw this one. I'm not sure where else is the actions-setup-openshift used. GitHub updated the ubuntu-latest to be based on 22.04 since the 1st of december. Unfortunately the action won't work in this version of ubuntu. Need to change the runs-on declaration to `ubuntu-20.04` that should fix it.

Many thanks!

PR #1111, opened 2022-12-14
Add property for ingressClassName

Sgitario:
Relates to https://github.com/quarkusio/quarkus/issues/29858


annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/decorator/AddIngressDecorator.java
```diff
@@ -54,13 +55,18 @@ public void visit(KubernetesListBuilder list) {
...
+    IngressFluent.SpecNested<IngressBuilder> ingressSpec = new IngressBuilder()
         .withNewMetadata()
         .withName(config.getName())
         .withLabels(allLabels)
         .endMetadata()
-        .withNewSpec()
-        .addNewRule()
+        .withNewSpec();
+
+    if (Strings.isNotNullOrEmpty(config.getIngress().getIngressClassName())) {
```
iocanel, line 65: Why not just use `withIngresClassName(config.getIngress().getIngressClassName())` it would save you the `if` and the clunky `IngressFluent` variable definition. If the target value is null it would be like it was never called.

Alternative, to avoid the `Ingressfluent` thing is to just use a visitor, but the first approach is far simpler.

iocanel (changes requested):
Added a minor change request.

annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/decorator/AddIngressDecorator.java
Sgitario: Good catch! My intention here was not to overwrite the ingress class name when it's set via a custom Ingress resource. 
I will split the Ingress decorator as done for the Route ingress decorator in https://github.com/dekorateio/dekorate/pull/1112.

Sgitario:
@iocanel I've already addressed the issue you spotted.

annotations/kubernetes-annotations/src/main/java/io/dekorate/kubernetes/decorator/AddIngressDecorator.java
Sgitario: It seems that the code was finally necessary. Otherwise, when there is no configuration, it adds the ingressClassName with value empty (""). 
I'm undoing this to the original changes.

iocanel: You can just use soemthing like: `withIngressClassName(Strings.isNotNullOrEmpty(config.getIngress().getIngressClassName()) ? config.getIngress().getIngressClassName() : null)`

Sgitario: I'm using `Strings.defaultIfEmpty(config.getIngress().getIngressClassName(), null)` instead. This seems to work fine.

iocanel (approved):

Issue #1116, opened 2022-12-21
Align service binding module with latest client release.


iocanel:
For `ServiceBinding` we manually define a service for `MappingProvider` that is used for deserialization. In the kubernetes-client 6.3.0 this has changed and we need to register the `KubernetesResource` instead.

PR #1117, opened 2022-12-21
Bump Kubernetes Client to 6.3.1

Sgitario:
Fix https://github.com/dekorateio/dekorate/issues/1116


core/src/main/java/io/dekorate/processor/AptReader.java
```diff
@@ -53,24 +56,6 @@ public class AptReader implements SessionReader, WithProject {
     this.processingEnv = processingEnv;
     this.inputFileIncludePattern = Pattern.compile(INPUT_FILE_INCLUDE_REGEX);
     this.inputFileExcludePattern = Pattern.compile(INPUT_FILE_EXCLUDE_REGEX);
-    hack();
-  }
-
-  private void hack() {
```
Sgitario, line 59: I'm not 100% sure that this is indeed no longer needed. (let's see what the CI says about this).

iocanel (approved):
LGTM

Issue #1124, opened 2023-01-12
Dekorate picks a random port for Route when having multiple container ports


Sgitario:
When having:

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: app
spec:
  ports:
    - name: https
      port: 443
      protocol: TCP
      targetPort: 8443
    - name: http
      port: 80
      protocol: TCP
      targetPort: 8080
  type: ClusterIP
```

When building the manifests, the generated Route in OpenShift sometimes picks HTTP and some othertimes uses HTTPS.

PR #1126, opened 2023-01-12
Set default `http` to `openshift.route.target-port` property

Sgitario:
And avoid using the problematic `getHttpPort(config)`. 
Fix https://github.com/dekorateio/dekorate/issues/1124


iocanel (approved):
LGTM

Issue #1120, opened 2023-01-11
Kind configuration sets the same nodeport for all the container ports


Sgitario:
When my deployment contains more than one container port, the generated Service looks like:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-app
spec:
  ports:
    - name: http
      nodePort: 30846
      port: 80
      protocol: TCP
      targetPort: 8080
    - name: https
      nodePort: 30846
      port: 443
      protocol: TCP
      targetPort: 8443
  selector:
    app.kubernetes.io/name: kind-with-defaults
    app.kubernetes.io/version: 0.1-SNAPSHOT
  type: NodePort
```

Note that the same nodePort is used for both HTTP and HTTPS which is not correct. A random one should be calculated.

PR #1127, opened 2023-01-12
Only set node-port for one service port in Kind and Minikube

Sgitario:
The rules are:
- If there is only one port, we use it
- If a user sets a node-port in one and only one of the ports in `dekorate.[kind or minikube or kubernetes].ports, we use it.
- Otherwise, we try to use the named port "http" -> this is for backward compatibility. 

Fix https://github.com/dekorateio/dekorate/issues/1120


iocanel (approved):
I relluctantly approve as I find it slightly hard to follow.
More comments etc could possibly help.

Sgitario:
> I relluctantly approve as I find it slightly hard to follow. More comments etc could possibly help.

PR updated according to your comments.

PR #1137, opened 2023-01-25
Include only the fabric8 resources for model in the apt dependencies

Sgitario:
Relates to https://github.com/quarkusio/quarkus/pull/30480

### DIFF
--- a/annotations/certmanager-annotations/pom.xml
+++ b/annotations/certmanager-annotations/pom.xml
@@ -196,8 +196,8 @@ limitations under the License.
                     <include>io/dekorate/certmanager/manifest/**</include>
                     <include>io/dekorate/certmanager/model/**</include>
                     <include>io/dekorate/certmanager/processor/**</include>
-                    <include>META-INF/services/io.fabric8.*</include>
                     <include>META-INF/services/io.dekorate.*</include>
+                    <include>META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource</include>
                     <include>META-INF/MANIFEST.MF</include>
                   </includes>
                 </filter>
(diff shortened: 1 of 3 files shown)


iocanel (changes requested):
We can't completely remove these lines, and still keep around: https://github.com/dekorateio/dekorate/blob/main/annotations/servicebinding-annotations/src/main/resources/META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource

We either need to remove both, or keep an entry like:
```xml
<include>META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource</include>
```

I am not sure however, if this is something we still use or it need also go away. Waiting for @manusa to clarify.

Sgitario:
> We can't completely remove these lines, and still keep around: https://github.com/dekorateio/dekorate/blob/main/annotations/servicebinding-annotations/src/main/resources/META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource
> 
> We either need to remove both, or keep an entry like:
> 
> ```
> <include>META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource</include>
> ```
> 
> I am not sure however, if this is something we still use or it need also go away. Waiting for @manusa to clarify.

Ups, sorry, my bad. I've already corrected the pull request. 
Please, @manusa confirm you're ok with these changes too.

manusa:
This is used by the `KubernetesDeserializer` to detect the different model types

https://github.com/fabric8io/kubernetes-client/blob/97f53c71d275e7544c7d757cc0fecb3fb0889109/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java#L194

https://github.com/fabric8io/kubernetes-client/pull/4511

manusa (approved):
Changes seem fine, thx

iocanel (approved):
lgtm

PR #1149, opened 2023-02-02
Fix decorator ordering with topological sort

iocanel:


Sgitario:
I was thinking of using a similar algorithm :) 
Let's fix the failures in the CI and we will know if it worked when running the downstream job in main

iocanel:
There are still multiple issues with this PR:
- The implementation is wrong (there are cases where it produces out of order decorators)
- Some decorators have cyclic dependencies.
- Some of the tests are not isolated.

It will keep me busy for the day.

iocanel:
#1150 seems to be in better shape than this one, closing.

PR #1150, opened 2023-02-03
Change dekorator ordering algorithm to use the topological sort algorithm

Sgitario:


core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    return sort(unordered.values());
+  }
+
+  /**
+   * Sort nodes with dependencies using a Graph Topological algorithm plus the visitor pattern to avoid having cycles and
+   * to gain better performance.
+   */
+  private static <T extends Comparable> List<T> sort(Collection<Node<T>> items) {
```
iocanel, line 95: While the method is parameterized, internally it used `Object` which feesl a bit off.

Either use an additional paramter for the node dependency type, or don't use parameters at all and just go with `Decorator` and `Class`.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -0,0 +1,225 @@
...
+          }
+        }
+      }
+    }
+
+    return sort(unordered.values());
+  }
+
+  /**
+   * Sort nodes with dependencies using a Graph Topological algorithm plus the visitor pattern to avoid having cycles and
```
iocanel, line 92: Can your provide more details about the algorithm used ?

iocanel (changes requested):

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: Yes, I'm planning to refactor and clean the current code.

core/src/main/java/io/dekorate/kubernetes/decorator/ApplyArgsDecorator.java
```diff
@@ -42,6 +42,6 @@ public void andThenVisit(ContainerFluent container) {
 
   public Class<? extends Decorator>[] after() {
     return new Class[] { ResourceProvidingDecorator.class, ApplyApplicationContainerDecorator.class,
-        AddSidecarDecorator.class };
+        AddSidecarDecorator.class, ApplyCommandDecorator.class };
```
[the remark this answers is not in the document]
Sgitario: Without this declaration, one Quarkus test was failing. But will revisit this later to ensure whether is necessary or not.

annotations/kind-annotations/src/main/java/io/dekorate/kind/manifest/KindManifestGenerator.java
```diff
@@ -33,21 +34,16 @@ public String getKey() {
...
-    if (config.getPorts().length > 0) {
-      resourceRegistry.decorate(group, new AddServiceResourceDecorator(config));
-    }
-  }
-
   @Override
   public void generate(KubernetesConfig kubernetesConfig) {
     initializeRegistry(kubernetesConfig);
-    addDecorators(KIND, kubernetesConfig);
+    Optional<KindConfig> kindConfig = configurationRegistry.get(KindConfig.class);
```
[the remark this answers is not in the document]
Sgitario: The problem here was that we were creating two decorators for the ImagePullPolicy, both for the kind group. One with the wrong imagePullPolicy and the other with the correct value. 
This change will ensure that only one decorator with the correct value is triggered.

annotations/kind-annotations/src/main/java/io/dekorate/kind/manifest/KindManifestGenerator.java
```diff
@@ -33,21 +34,16 @@ public String getKey() {
     return KIND;
   }
 
-  @Override
```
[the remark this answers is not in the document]
Sgitario: The ports are being properly addressed in the super implementation, we didn't need to do this here. (The same decorators were registered twice).

Sgitario:
TODO:
- Refactor current solution
- Raise exceptions when a cycle is detected

core/src/main/java/io/dekorate/kubernetes/decorator/ApplyArgsDecorator.java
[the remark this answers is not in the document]
Sgitario: Reverted

Sgitario:
> TODO:
> 
> * Refactor current solution
> * Raise exceptions when a cycle is detected

@iocanel I've already refactored the current solution and also change the cycle logic to raise exceptions instead. 
The only cycle detected in Dekorate was fixed by https://github.com/dekorateio/dekorate/pull/1150/files#diff-6842f233d53d56ced32c89058673e0286f831f6b9ff881de9269483577fd2152R96. 

There is also a cycle exception in Quarkus that we would need to address in Quarkus:

```
Build step io.quarkus.kubernetes.deployment.KubernetesProcessor#build threw an exception: java.lang.RuntimeException: Cycle detected when ordering decorators: 
- io.quarkus.kubernetes.deployment.RemoveOptionalFromSecretEnvSourceDecorator
- io.quarkus.kubernetes.deployment.RemoveOptionalFromSecretKeySelectorDecorator
- io.quarkus.kubernetes.deployment.RemoveOptionalFromConfigMapKeySelectorDecorator
- io.quarkus.kubernetes.deployment.RemoveOptionalFromConfigMapEnvSourceDecorator

```

annotations/kind-annotations/src/main/java/io/dekorate/kind/manifest/KindManifestGenerator.java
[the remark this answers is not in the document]
iocanel: Makes sense

annotations/kind-annotations/src/main/java/io/dekorate/kind/manifest/KindManifestGenerator.java
```diff
@@ -33,21 +34,16 @@ public String getKey() {
...
-    }
-  }
-
   @Override
   public void generate(KubernetesConfig kubernetesConfig) {
     initializeRegistry(kubernetesConfig);
-    addDecorators(KIND, kubernetesConfig);
+    Optional<KindConfig> kindConfig = configurationRegistry.get(KindConfig.class);
+
+    super.addDecorators(KIND, new KubernetesConfigBuilder(kubernetesConfig)
```
iocanel, line 42: Why are you using `super.addDecorators`?

Sgitario: do you mean the `super.` appendix? We don't need it, we can simply use `addDecorators`

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: This was based on an agnostic implementation I had in another project. I've already changed to use the Decorator type directly.

annotations/kind-annotations/src/main/java/io/dekorate/kind/manifest/KindManifestGenerator.java
Sgitario: If it was that, I've already removed the `super.`.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -0,0 +1,218 @@
...
+    // First, we loop over each node in order of insertion.
+    ArrayDeque<Node> back = new ArrayDeque<>();
+    back.addAll(items);
+
+    while (!back.isEmpty()) {
+      // Get the first node.
+      Node node = back.pop();
+      Class identity = node.value;
+      if (isVerbose()) {
+        LOGGER.info("[sort] Sort " + identity);
```
iocanel, line 110: This log entry doesn't provide any meaningful information.
This could be something like:
- visiting
- checking 
- collapsed into one of the log entries belllow (this would probably provide a cleaner output).

core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -0,0 +1,218 @@
...
+        cycle.clear();
+        visited.add(identity);
+        ordered.addAll(node.references);
+      } else {
+        boolean ready = true;
+        for (Class d : dependants) {
+          if (!visited.contains(d)) {
+            ready = false;
+            if (isVerbose()) {
+              LOGGER.info("[sort] " + identity + " is not ready because it needs " + d);
```
iocanel, line 123: Continuing my previous comment. This entry could be merged with the one above so that we have one line per node poped, or something like this.

iocanel (approved):
LGTM.

Added some optional remarks regarding the log.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: Yes, this log does not add much value. I've deleted it.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: I keep this trace here because I deleted the previous one.

Sgitario:
Failure is unrelated.

PR #1154, opened 2023-02-09
Guarantee initial order for decorators to sort

Sgitario:
Plus, add some test coverage with basic scenarios.

### DIFF
--- a/core/src/main/java/io/dekorate/utils/TopologicalSort.java
+++ b/core/src/main/java/io/dekorate/utils/TopologicalSort.java
@@ -11,6 +11,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 
 import io.dekorate.Logger;
 import io.dekorate.LoggerFactory;
@@ -32,7 +33,8 @@ public final class TopologicalSort {
    */
   public static List<Decorator> sortDecorators(Collection<Decorator> items) {
     Collection<Node> nodes = adaptToNodes(items);
-    return sortNodes(nodes);
+    // Order the nodes by the decorator class to guarantee the initial order.
+    return sortNodes(new TreeSet<>(nodes));
   }
 
   /**
@@ -169,7 +171,7 @@ public final class TopologicalSort {
   /**
    * This class is internally used to represent a node of decorators and its dependencies.
    */
-  private static class Node {
+  private static class Node implements Comparable<Node> {
     private final Class value;
 
     private List<Decorator> references = new ArrayList<>();
@@ -182,6 +184,11 @@ public final class TopologicalSort {
     @Override
     public String toString() {
       return value.getName();
+    }
+
+    @Override
+    public int compareTo(Node o) {
+      return value.getName().compareTo(o.value.getName());
     }
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -183,6 +185,11 @@ public Node(Class value) {
     public String toString() {
       return value.getName();
     }
+
+    @Override
+    public int compareTo(Node o) {
+      return value.getName().compareTo(o.value.toString());
```
iocanel, line 191: This seems awkward.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
```diff
@@ -32,7 +33,8 @@ private TopologicalSort() {
    */
   public static List<Decorator> sortDecorators(Collection<Decorator> items) {
     Collection<Node> nodes = adaptToNodes(items);
-    return sortNodes(nodes);
+    // Order the nodes by the decorator class to guarantee the initial order.
+    return sortNodes(new TreeSet<>(nodes));
```
iocanel, line 37: It's not clear to me what you mean by initial order.

iocanel (changes requested):
The compareTo seems awkward so it either needs to change or some comments explaining why.
The TreeSet part is also not clear to me.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: Let's discuss this in the other comment.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: The idea is to get a determinist order if we always get the same input. For example, if our input is a set of three decorators with no dependencies on each other, the final result order of these three decorators might vary in different executions.

The idea I had to address this situation is to introduce an initial order based on the decorator class name (it can be any other).

core/src/main/java/io/dekorate/utils/TopologicalSort.java
iocanel: Sorry, I should be more specific. I mean that comparing `getName()` to `toString()` seems very weird.

Sgitario: Very weird and certainly wrong :p 
Will fix it.
Thanks for spotting!

core/src/main/java/io/dekorate/utils/TopologicalSort.java
iocanel: > The idea is to get a determinist order if we always get the same input. For example, if our input is a set of three decorators with no dependencies on each other, the final result order of these three decorators might vary in different executions.

The order of decorators should not be important to us as long as they are topologically sorted. Even if it's possible the get different results that both satisfy the constraints, the end result (the generated json/yaml files) should be deterministic. If the end result is not deterministic, then we probably have not properly expressed those constraints and we should focus on that.

TDLR: I don't mind having the decorators in a predictable order however, we shouldn't have to,  are we sure we are dealing with the issue at the right level?

iocanel (approved):
LGTM as long as we fix the toString thingy.

core/src/main/java/io/dekorate/utils/TopologicalSort.java
Sgitario: Fixed

Issue #1158, opened 2023-02-27
Certmanager annotation config fails missing name


aureamunoz:
The annotation config for certmanager module doesnt work.

Use the [Spring boot cert manager example](https://github.com/dekorateio/dekorate/blob/6473b699627c2eecf1a45c4263555bc52a4f6bc6/examples/spring-boot-with-certmanager-example/src/main/java/io/dekorate/example/Main.java#L33-L32) for reproducing:

1 - Comment out the cert manager configuration in the application.properties file: 
```
#dekorate.certificate.secret-name=tls-secret
#dekorate.certificate.dnsNames=kubernetes-example.com,localhost
#dekorate.certificate.self-signed.enabled=true
#dekorate.certificate.subject.organizations=Dekorate,Community
#dekorate.certificate.duration=2160h0m0s
#dekorate.certificate.renewBefore=360h0m0s
#dekorate.certificate.privateKey.algorithm=RSA
#dekorate.certificate.privateKey.encoding=PKCS8
#dekorate.certificate.privateKey.size=2048
#dekorate.certificate.keystores.pkcs12.create=true
#dekorate.certificate.keystores.pkcs12.passwordSecretRef.name=pkcs12-pass
#dekorate.certificate.keystores.pkcs12.passwordSecretRef.key=password
#dekorate.certificate.usages=server auth,client auth
```

2-Add the configuration using annotations, in [this class](https://github.com/dekorateio/dekorate/blob/6473b699627c2eecf1a45c4263555bc52a4f6bc6/examples/spring-boot-with-certmanager-example/src/main/java/io/dekorate/example/Main.java#L29):

```

@Certificate(name = "cert-name", secretName = "tls-secret", selfSigned = @SelfSigned(enabled = true), usages={"server auth", "client auth"}, dnsNames={"kubernetes-example.com","localhost"}, organizations={"Dekorate", "Community"}, duration="2160h0m0s", renewBefore="360h0m0s",
  privateKey = @CertificatePrivateKey(algorithm= PrivateKeyAlgorithm.RSA, encoding = PrivateKeyEncoding.PKCS8, size = 2048), keystores= @CertificateKeystores(pkcs12 = @CertificateKeystore(create = true,passwordSecretRef = @LocalObjectReference(name ="pkcs12-pass", key="password" ))))
```

3 - Build the project from the root: 
`mvn clean install`

The build fails with: 

```
[...]

PR #1159, opened 2023-02-28
Cert Manager: Fix extension to get config from annotations

Sgitario:
Fix https://github.com/dekorateio/dekorate/issues/1158


aureamunoz (approved):
Great! LGTM Thank you

PR #1166, opened 2023-03-21
Fix template metadata is not generated in Deployment resources

Sgitario:
Relates to CI failures: https://github.com/quarkusio/quarkus/runs/12159739217#user-content-test-failure-io.quarkus.it.kubernetes.kuberneteswithmetricscustomabsolutetest-1


core/src/main/java/io/dekorate/kubernetes/decorator/DeploymentResourceFactory.java
```diff
@@ -39,9 +40,10 @@ public HasMetadata create(AbstractKubernetesManifestGenerator<?> generator, Base
         .endMetadata()
         .withNewSpec()
         .withReplicas(1)
-        .withNewTemplate()
```
iocanel, line 42: I disagree with this change. 

Our decorators should work regardless of the structure of the initial resource (users are allowed to bring their own resource and can't possibly know the decorate internals).

When we have a decorator that requires the exististence of some objects, we should then ensure that these objects exists (usually by also adding the decorators that create them).

So, instead of creating the template and metadata here, we should be using the decorators that do so.

PR #1167, opened 2023-03-21
fix: before() for AddMetadataToTemplateDecorator

iocanel:


Sgitario (approved):
I did check that with your changes, the failing tests are now working fine.
Many thanks!

PR #1171, opened 2023-03-22
chore: delete tmp files that back streams when stream is closed

iocanel:
The pull request cleans up temporary files that are used to support Streams.
Closing happens when stream closes which render the File no longer accessible.


iocanel:
cc @kabir

Sgitario (approved):

PR #1189, opened 2023-04-19
Fix decorator cycle issue with Cert-Manager and OpenShift

Sgitario:
Relates to https://github.com/quarkiverse/quarkus-cert-manager/pull/39#issue-1674707504


annotations/certmanager-annotations/src/main/java/io/dekorate/certmanager/decorator/AddCertificateResourceDecorator.java
```diff
@@ -26,26 +25,25 @@
...
 
   @Override
   public void visit(KubernetesListBuilder list) {
-    HasMetadata meta = getMandatoryDeploymentHasMetadata(list, ANY);
     CertificateBuilder builder = new CertificateBuilder();
 
     // metadata
     builder.withNewMetadata()
-        .withName(getName(meta))
-        .withNamespace(meta.getMetadata().getNamespace())
```
Sgitario, line 48: Updating the namespace was not necessary as there is a decorator that populates all the resources with the correct namespace.

annotations/certmanager-annotations/src/main/java/io/dekorate/certmanager/manifest/CertificateResourceGenerator.java
```diff
@@ -56,27 +58,31 @@ public void generate(CertificateConfig config) {
     LOGGER.info("Processing certificate config.");
     validate(config);
 
+    String defaultName = configurationRegistry.get(BaseConfig.class).map(BaseConfig::getName)
+        .orElse(config.getProject().getBuildInfo().getName());
+    String name = defaultIfEmpty(config.getName(), defaultName);
```
Sgitario, line 63: This is the important fix: Instead of getting the name from the generated deployment resource, it will get it from the BaseConfig.
I've confirmed that this works fine for the kubernetes and openshift extensions.

iocanel (approved):
lgtm

PR #1192, opened 2023-05-10
Helm: Add property to configure the separator for profile values files

Sgitario:
Related to https://github.com/quarkiverse/quarkus-helm/pull/246


iocanel (approved):
LGMT

PR #1193, opened 2023-05-10
chore: centralize handling of container decorator order

iocanel:
The pull request move containers dekorator order handling (before / after) methods to the `ApplicationContainerDecorator` class. Also, adds ApplicatonContainerDecorators to run after `AddInitContainerDecorator`.


iocanel:
@Sgitario: `AddInitContainerDecorator` was not ordered correctly. With this change all of our existing decorators that modify containers (extending ApplicationContainerDecorator) should run after `AddInitContainerDecorator`. 

So, theoretically we should be able to change image, command, args, env etc (for init containers) without introducing new decorators.

Sgitario (approved):
Awesome!

PR #1194, opened 2023-05-10
Helm: Bind port numbers in Helm values file

Sgitario:
The port number should be inline in several locations: deployment resources, service and HTTP probes.


iocanel (approved):
LGTM

PR #1196, opened 2023-05-17
Support overwrite values from existing init-containers

Sgitario:
Related to https://github.com/quarkusio/quarkus/issues/33097#issuecomment-1533019528

### DIFF
--- a/core/src/main/java/io/dekorate/kubernetes/decorator/AddInitContainerDecorator.java
+++ b/core/src/main/java/io/dekorate/kubernetes/decorator/AddInitContainerDecorator.java
@@ -18,8 +18,10 @@ package io.dekorate.kubernetes.decorator;
 import io.dekorate.doc.Description;
 import io.dekorate.kubernetes.adapter.ContainerAdapter;
 import io.dekorate.kubernetes.config.Container;
+import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PodSpecBuilder;
+import io.fabric8.kubernetes.api.model.PodSpecFluent;
 
 /**
  * A decorator that adds an init container to a pod template.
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/dekorate/kubernetes/decorator/AddInitContainerDecorator.java
```diff
@@ -36,6 +38,55 @@ public AddInitContainerDecorator(String deployment, Container container) {
...
+  private void update(PodSpecBuilder podSpec, io.fabric8.kubernetes.api.model.Container resource) {
+    PodSpecFluent.InitContainersNested<PodSpecBuilder> matching = podSpec
+        .editMatchingInitContainer(this::existsContainerByName);
+
+    if (resource.getImage() != null) {
+      matching.withImage(resource.getImage());
+    }
+
+    if (resource.getWorkingDir() != null) {
+      matching.withWorkingDir(resource.getWorkingDir());
```
iocanel, line 62: Most String values in the annotation default to empty String. I don't recall by heart how empty Strings translate in the generated Pojos. This highlights an issue: 
- The correctness of the this code depends on `what sundrio does for generating Pojos`.

Looking deeper I think that the correctness also depends on:
- Kubernetes model default values.
- ContainerAdapter  logic.

This is why I am always a little bit sceptic when it comes to decorators that have such a large surface area.

Given, that this is clearly an improvement over what we have and lacking a better proposal I'll let it pass, but we  need to be aware of the dangers in case it backfires.

tests/feat-overwrite-init-containers/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+dekorate.kubernetes.init-containers[0].name=foo
```
iocanel, line 1: Would it be possible to switch to a more realistic scenario, like having the user provide their own resources containing an  init  container and using property configuration to change things like env, image etc?

iocanel (changes requested):
It's clearly an improvement over what we have.
I'd like the following:
- apply the same principal for sidecars too.
- change to a more real world example in the integration test.

tests/feat-overwrite-init-containers/src/main/resources/application.properties
Sgitario: This is a very fair request and indeed the first attempt was to replace the init container image configuration. The problem that I encountered is that the order of the init container was not respected. For example:

```
dekorate.kubernetes.init-containers[0].name=foo
dekorate.kubernetes.init-containers[0].command=sh
dekorate.kubernetes.init-containers[1].name=foo
dekorate.kubernetes.init-containers[1].command=sh updated
```

In this example, the `init-containers[1]` was applied first and then `init-containers[0]` which I think it's wrong. wdyt?

core/src/main/java/io/dekorate/kubernetes/decorator/AddInitContainerDecorator.java
Sgitario: Yes, I see your point and I agree with you about how dangerous/wrong this code could be. The root cause is that annotations default strings to empty and the properties default to null :/

tests/feat-overwrite-init-containers/src/main/resources/application.properties
iocanel: I wasmostly asking to use an existing resource vs defining both init containers via props. Anyway,  this is no biggie. So let's move on.

iocanel (commented):
lgtm

PR #1200, opened 2023-05-24
Preserve the order of decorators by declaration

Sgitario:


iocanel (approved):
lgtm

Sgitario:
The failures on the snowdrop examples are expected until we can upgrade Dekorate to the latest version. Merging.

PR #1206, opened 2023-06-01
[2.11] Revert "fix: route port"

Sgitario:
This reverts commit 21ef3c382db86394fed765eabb3b30eb8609baf1.

### DIFF
--- a/annotations/openshift-annotations/src/main/java/io/dekorate/openshift/decorator/AddRouteDecorator.java
+++ b/annotations/openshift-annotations/src/main/java/io/dekorate/openshift/decorator/AddRouteDecorator.java
@@ -62,7 +62,7 @@ public class AddRouteDecorator extends ResourceProvidingDecorator<KubernetesList
         .withName(config.getName())
         .endTo()
         .withNewPort()
-        .withNewTargetPort(port.getContainerPort())
+        .withNewTargetPort(port.getHostPort())
         .endPort()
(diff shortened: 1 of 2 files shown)


annotations/openshift-annotations/src/main/java/io/dekorate/openshift/decorator/AddRouteDecorator.java
```diff
@@ -62,7 +62,7 @@ public void visit(KubernetesListBuilder list) {
         .withName(config.getName())
         .endTo()
         .withNewPort()
-        .withNewTargetPort(port.getContainerPort())
+        .withNewTargetPort(port.getHostPort())
```
iocanel, line 65:
```suggestion
        //A service should be consumed without requiring consumers to know the container port.
        //In the same spirit the route should not refer to container port (as the mapping needs to happen on Service level).
        .withNewTargetPort(port.getHostPort())
```

iocanel (approved):
LGTM

Addded a suggestion: A comment that hopefully will prevent us from breaking this again.

Sgitario:
We were wrong about reverting this change. From OpenShift docs:
`targetPort is the target port on pods that is selected by the service that this route points to.`

So, we should keep using the container port, not the host-port.

Sgitario:
By the way, this was spotted by one test we have in our CI :)

PR #1210, opened 2023-06-12
Ensure always pickup the same port

Sgitario:
If the container ports are generated in a different order, the port is pickup in a different order. 

With this change, the logic to pick up the http to use is consistent.

I saw this issue when generating a Quarkus application. It randomly picks up the HTTP or the HTTPS in the generated Health probes.


iocanel (approved):
lgtm

PR #1214, opened 2023-06-14
fix: resolution of git remotes

Sgitario:
Once the first git remote is found, we iterate over all the values and overwrite the values.

For example, having:

```
[remote "origin"]
	url = git@github.com:Sgitario/dekorate.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[remote "upstream"]
	url = git@github.com:dekorateio/dekorate.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
[remote "auri"]
	url = https://github.com/aureamunoz/dekorate
	fetch = +refs/heads/*:refs/remotes/auri/*
```

Will produce a map of only an entry with:

```
'remote "origin"': https://github.com/aureamunoz/dekorate
```

Because https://github.com/aureamunoz/dekorate, it's the last value.

Also, the logic about remotes need to be aligned with this in the TektonManifestGenerator.


iocanel (approved):
lgtm

PR #1216, opened 2023-06-16
fix: integration with Tekton

Sgitario:
The integration with Tekton was broken for the following reasons:
- The service account integration was totally broken: so, neither private container registry could be used, nor the k8s manifests could be installed
- The kaniko image used to build and push the images was not working (one missing parameter and a wrong parameter related to the context)
- The extension was not supporting a multi-module project
- The image-pull-secret property was not taking into account: I had to revert https://github.com/dekorateio/dekorate/commit/1bcf751b08fca189fcb2f86d75a946da7dc09831
- Updated documentation and CI

All above have been fixed and addressed.

### DIFF
--- a/.github/workflows/integration-tests-with-dockerio.yml
+++ b/.github/workflows/integration-tests-with-dockerio.yml
@@ -21,7 +21,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
 
 jobs:
   cache:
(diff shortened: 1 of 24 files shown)


iocanel (approved):
LGTM
For future reference, I would prefer individaul pull requests per problem fixed, as this make it a bit hard to review.

Sgitario:
> LGTM For future reference, I would prefer individaul pull requests per problem fixed, as this make it a bit hard to review.

To ease the review, there are separate commits within this pull request. 
Pull request should address or handle one issue, and I did it when it made sense (see https://github.com/dekorateio/dekorate/pull/1214, https://github.com/dekorateio/dekorate/pull/1213, https://github.com/dekorateio/dekorate/pull/1212, https://github.com/dekorateio/dekorate/pull/1211).

PR #1218, opened 2023-06-19
Tekton: Migrate from PipelineResources

Sgitario:


core/src/main/java/io/dekorate/kubernetes/decorator/AddDockerConfigJsonSecretDecorator.java
```diff
@@ -68,16 +68,21 @@ public AddDockerConfigJsonSecretDecorator(String name, Path existingConfigJson,
...
-    Secret secret = new SecretBuilder()
+    SecretBuilder secret = new SecretBuilder()
         .withNewMetadata()
         .withName(name)
         .withAnnotations(this.annotations)
         .endMetadata()
         .withType(TYPE)
-        .addToData(DOT_DOCKER_CONFIG_JSON, this.content)
-        .build();
+        .addToData(DOT_DOCKER_CONFIG_JSON, this.content);
```
iocanel, line 77: Why don't you just make that key configuralbe? It will save you the `doUpate` and extending thing.
Alterntatively, you can just create a decorator that applies the name of they key.

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/decorator/AddDockerConfigJsonSecretDecorator.java
```diff
@@ -0,0 +1,60 @@
...
+import io.fabric8.kubernetes.api.model.SecretBuilder;
+
+/**
+ * Special version of AddDockerConfigJsonSecretDecorator for Tekton where some tasks need the docker config file
+ * in .dockerconfigjson and some other tasks/steps require it in config.json.
+ *
+ * For example, the kaniko image explicitly requires the config.json file. See more in
+ * https://artifacthub.io/packages/tekton-task/tekton-tasks/kaniko?modal=manifest.
+ */
+public class AddDockerConfigJsonSecretDecorator extends io.dekorate.kubernetes.decorator.AddDockerConfigJsonSecretDecorator {
```
iocanel, line 32: We shouldn't need to extend this class just to use a different `file` name .

iocanel (commented):
@Sgitario I tried reviewing the PR but I am missing the context.
It's not clear to me what you are trying to do and why.

Some rough examples:
- Why do we need a clone step ?
- Why in some areas are replacing ouput references for had coded values ?

Note, that I have't touched tekton for nearly two years now, so I am bit rusty.

Sgitario:
> @Sgitario I tried reviewing the PR but I am missing the context. It's not clear to me what you are trying to do and why.
> 
> Some rough examples:
> 
> * Why do we need a clone step ?

We need a git clone step, because the PipelineResources of type git is removed in Tekton 0.47. For the migration, I did what was described in https://tekton.dev/docs/pipelines/resources/ which says to replace the pipelineresources to params and use the `git-clone` task (see the image). 

> * Why in some areas are replacing ouput references for had coded values ?
> 

As far as I see, there are no hard-coded values, but default values that users can modify if needed. I don't think this has changed by these changes. But before these properties were provided via resources inputs/outputs, and now via params. 

> Note, that I have't touched tekton for nearly two years now, so I am bit rusty.

iocanel:
Thanks @Sgitario much clearer now.

Sgitario:
> Thanks @Sgitario much clearer now.

No problem!

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/step/KanikoBuildStep.java
```diff
@@ -24,20 +24,15 @@
 public class KanikoBuildStep extends ImageBuildStep<KanikoBuildStep> {
 
   public static final String DOCKERFILE_FORMAT = "--dockerfile=%s";
-  public static final String DOCKERFILE_ARG = "--dockerfile=$(inputs.params." + PATH_TO_DOCKERFILE_PARAM_NAME + ")";
   public static final String CONTEXT_FORMAT = "--context=%s";
-  public static final String CONTEXT_ARG = "--context=$(params." + PATH_TO_CONTEXT_PARAM_NAME + ")";
   public static final String WORKSPACE_SOURCE = "/workspace/source";
   public static final String FORCE_ARG = "--force";
   public static final String INSECURE_ARG = "--insecure";
-  public static final String IMAGE_DESTINATION_ARG = "--destination=$(resources.outputs.image.url)";
```
iocanel, line 33: I see in many places replacing `$(resources.outputs.image.url)` it's not clear to me why we do this.
Is it because the `output` resource is no longer available ? Is it something else?

core/src/main/java/io/dekorate/kubernetes/decorator/AddDockerConfigJsonSecretDecorator.java
Sgitario: Because we need these two keys. The `config.json` is for the step and the `.dockerconfigjson` one is the standard for Kubernetes image pull secret. 
I wrote a commend in the class that explained why it was required.

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/decorator/AddDockerConfigJsonSecretDecorator.java
Sgitario: Same comment than in https://github.com/dekorateio/dekorate/pull/1218#discussion_r1236705012

annotations/tekton-annotations/src/main/java/io/dekorate/tekton/step/KanikoBuildStep.java
Sgitario: The `$(resources...` is deprecated (and removed in Tekton v0.47 which is the version that the Fabric8 Kubernetes Client 6.7.2 uses). The replacement is to use the params.

iocanel (approved):
LGTM

PR #1222, opened 2023-06-21
feat: support deployment strategy rolling update in DeploymentConfig

Sgitario:
Related to https://github.com/quarkusio/quarkus/pull/34179


annotations/openshift-annotations/src/main/java/io/dekorate/openshift/manifest/OpenshiftManifestGenerator.java
```diff
@@ -140,14 +142,23 @@ protected void addDecorators(String group, OpenshiftConfig config) {
...
+    boolean httpsPreferred = vcsConfig.map(VcsConfig::isHttpsPreferred).orElse(false);
 
     String vcsUrl = project.getScmInfo() != null && Strings.isNotNullOrEmpty(project.getScmInfo().getRemote().get(Git.ORIGIN))
-        ? Git.getRemoteUrl(project.getRoot(), remote, httpsPrefered).orElse(Labels.UNKNOWN)
+        ? Git.getRemoteUrl(project.getRoot(), remote, httpsPreferred).orElse(Labels.UNKNOWN)
         : Labels.UNKNOWN;
 
     resourceRegistry.decorate(group, new AddVcsUrlAnnotationDecorator(config.getName(), OpenshiftAnnotations.VCS_URL, vcsUrl));
     resourceRegistry.decorate(group, new AddCommitIdAnnotationDecorator());
+    if (DeploymentResourceFactory.KIND.equalsIgnoreCase(config.getDeploymentKind())) {
```
iocanel, line 153: This assumes that deployment kind is either `Deployment` or `DeploymentConfig` which is not accurate as we might have `StatefulSet` etc.

Eitherway, explicit checking of kind is not really needed as this is the responsibility of the decorator itself. In other words you can just add both and each will be applied where it makes sense.

Personally, I am ok eitherways (with or without checking deployment kind), but since the existing if is not accurate as described above, we might be better of without it.

Sgitario: The `StatefulSet` resource does not have the rolling update or strategy configuration, that's why it's ignored. 
About creating both decorators at the same time, I know it will work, but I would prefer creating only one depending on the deployment target kind selected.

iocanel: It doesn't really matter if `StatefulSet` is supported or not. The code assumes that if its not a `Deployment` its a `DeploymentConfig` which is inacurrate. Please either address that or remove the if / else completely.

iocanel (changes requested):
There is still an iterm that needs to be addressed.

annotations/openshift-annotations/src/main/java/io/dekorate/openshift/manifest/OpenshiftManifestGenerator.java
Sgitario: I know what you mean now. PR updated.

iocanel (approved):
lgtm

PR #1235, opened 2023-08-14
Release 4.0.0 ?

iocanel:
While I am on PTO I am creating this pull request mostly to lay out my thoughts on how we should proceed regarding the next release....

Since we upgraded main to the kubernetes-client model to 6.8.0 that brings massive changes to the API, the next release from main *needs* to bump the major version to 4. 

Fixes for 3.x version should happen from 3.x branch.

We may hold this around for a while if we are planning for other breaking changes that should be included in 4.0.0.

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,4 +1,4 @@
 name: Dekorate
 release:
-  current-version: 3.7.1
-  next-version: 3.7-SNAPSHOT
+  current-version: 4.0.0
+  next-version: 999-SNAPSHOT


metacosm:
Problem is, people are asking for the 6.8 upgrade in JOSDK. While we could do it, I would rather make sure first that Quarkus can be updated to it first and this release is holding that upgrade.

Sgitario:
> Problem is, people are asking for the 6.8 upgrade in JOSDK. While we could do it, I would rather make sure first that Quarkus can be updated to it first and this release is holding that upgrade. So, while there's no urgency to release, it should also happen fairly quickly, imo, like weeks rather than months.

Added the label `downstream-check` to verify these changes with Quarkus.

iocanel:
> > Problem is, people are asking for the 6.8 upgrade in JOSDK. While we could do it, I would rather make sure first that Quarkus can be updated to it first and this release is holding that upgrade. So, while there's no urgency to release, it should also happen fairly quickly, imo, like weeks rather than months.
> 
> Added the label `downstream-check` to verify these changes with Quarkus.

I don't think its possible for the downstream check to pass as it needs at least an additional verssion bump for the kubernetes client.

metacosm:
Can we please get an update on this, @iocanel?

metacosm:
So will Quarkus be updated to this new 4.0.0 version, because that's the real blocker, if I understand things properly?

PR #1262, opened 2023-10-24
Fix openshift annotation for link to version control source

indiealexh:
Fix openshift annotation for link to version control source

As per https://docs.openshift.com/container-platform/4.13/applications/odc-viewing-application-composition-using-topology-view.html#odc-labels-and-annotations-used-for-topology-view_viewing-application-composition-using-topology-view should be vcs-uri not vcs-url


iocanel:
@indiealexh: Thank you for this.

PR #1296, opened 2024-06-28
Fix Label not using correct kinds

xstefank:
When I tried something like `new AddLabelDecorator(name, label.getKey(), label.getValue(), ROUTE)` it didn't work. I figured out this fix which fixed it for my case but feel free to point any issues since this is my first PR to this project.


iocanel (approved):
Great catch!

PR #1316, opened 2026-01-27
fix: vulnerability when parsing untrusted XML via isXmlEqualTo 

aureamunoz:
Fixes [CVE-2026-24400](https://github.com/dekorateio/dekorate/security/dependabot/93)


metacosm (approved):
This is a dummy vulnerability, imo. No one is shipping assertj in code that runs in prod. Or, if they do, they deserve to be pown! 😅

geoand (approved):

